PR #313, opened 2025-10-24
fix(control): actually fix the bug causing CC not reaching 127

ArthurGibert:


chmanie (approved):
Alright! One comment would be to maybe clean up comments but let's get this out there!

PR #342, opened 2025-11-12
Start with running clock & save clock state

chmanie:
This PR changes the internal clock behavior so that when the device is booted up for the first time, the clock is running already (to counter the "it doesn't do anything" issue). After that the clock state is stored in memory.


ArthurGibert (approved):
Looks good, Unfortunately I was a little fast and cannot confirm if the clock is on by default.

I'll calibrate some units next week this will be confirmed then

PR #350, opened 2025-11-20
Add factory reset function to configurator

chmanie:
This PR adds a setting to the configurator to perform a factory reset for the device. This deletes all settings, params, scenes and storage but keeps the calibration data.


ArthurGibert (approved):
A scary one to approve but so far testing was good

PR #375, opened 2025-12-09
Add AD velocity 0 note off detection

ArthurGibert:


ArthurGibert:
Yeah that works, let's do that

chmanie (approved):
Great!

PR #380, opened 2025-12-10
Add Control X2 app

ArthurGibert:
Like the control app but the button sends CC instead of being a mute


ArthurGibert:
@chmanie this is re-based now and ready for merge after your review

Issue #411, opened 2026-01-30
Add some more exotic scales


rjsmith:
Add some more exotic scales to the current list.

Source: https://github.com/thorinside/soma/blob/main/soma.lua#L8-L29

chmanie:
I think I would close this in favor of user scales #500

PR #412, opened 2026-01-30
Add exotic scales from DistingNT Soma sequencer

rjsmith:
Added 11 more scales to the global list, ported from the Soma sequencer https://github.com/thorinside/soma/blob/main/soma.lua#L8-L29

refs ATOVproject/faderpunk#411

NB: Quote from Chris | ATOV:

> The list is limited to 16 currently as we only have 16 sort of distinguishable colors to display them. But we want to use the LEDs to actually show the included notes, when we do that we’re pretty much not limited (well, by 12 bit)


chmanie:
I'll close this in favor of #500. Arthur and I decided that having user scales provides more flexibility

PR #415, opened 2026-02-05
LFO+: Reset CV destination

ArthurGibert:
Add Reset CV destination to LFO+ and change documentation accordingly


configurator/src/components/ManualTab.tsx
```diff
@@ -845,7 +845,7 @@ const apps: ManualAppData[] = [
         fnDescription: "",
         fnPlusShiftTitle: "CV destination",
         fnPlusShiftDescription:
-          "Speed (yellow), phase (pink), amplitude (cyan)",
+          "Speed (yellow), phase (pink), amplitude (cyan), reset (pink)",
```
chmanie: Is it deliberate that phase and reset are both pink?

PR #432, opened 2026-02-19
Created a Scene page when holding the Scene button

ArthurGibert:
Scene button now flashed to the tempo

When holding the Scene button:
- the last recalled scene is displayed
- The tempo is displayed on fader 15 top LED
- The scale and tonic color are displayed on the top fader LEDs
- The scale is represented as a musical keyboard on channels 3-14 bottom LEDs (Dimmed out of scale, highlighted in the scale)


chmanie:
I have no idea which commits belong here, maybe you could clean this up?

PR #454, opened 2026-03-01
Add FP-Grids app

rjsmith:
A port of Emilie Gillet's renowned Mutable Instruments Grids topographic drum sequencer for the ATOV Faderpunk


chmanie:
I'm carefully closing this as #467 is there now? Let me know if anything from here is missing.

PR #455, opened 2026-03-02
Division param for clk div

ArthurGibert:
Add a param to select straight, tripplet or both types of division


chmanie (approved):
Very cool!

PR #473, opened 2026-03-06
fix(clock): improve clock out for consistent order

ArthurGibert:
Here is the context for this.

When I'm using a clock out with a long division (1 bar or up) to reset the sequencers in my system at regular interval. Sometimes the longer division would be sent before or after the 4ppqn clock pulse. That is causing the sequences to misalign in an unpredictable manner (which is not great when playing live.

This PR makes sure the longer divisions (1bar and up) are sent before the shorter ones.


chmanie:
Superseded by #491

PR #467, opened 2026-03-03
FP Grids

ArthurGibert:
A port of Emilie Gillet's renowned Mutable Instruments Grids topographic drum sequencer for the ATOV Faderpunk

PR #477, opened 2026-03-13
Alternative fixes for FP-Grids phasing and clock div issues

rjsmith:
Removed the new pattern generator code that double-stepped the Drum mode sequence. Instead, Drums Mode now uses a simple fixed clock division of 1/32nd notes to match the original firmware.  I don't think this mode needs ability to vary the clock division (making it more complex to use) .. there is a lot you can do with just moving the channel density sliders to vary the rhythmic pulse of the pattern.  

Euclidean mode uses the clock divider, that works just fine.   If it seems too complex that only one of three modes has a built-in clock divider, we could just remove it from Euclidean mode to match the OG firmware (there is always the FP's own Euclid app if anyone wants that). 

This MR also adds a not very carefully hidden Easter Egg mode, documented in the fp_grids.rs file.  🥁


ArthurGibert (approved):
I think it's ready to go now!

chmanie:
Closing this in favor of #467

PR #491, opened 2026-03-23
feat(faderpunk): add unified clock sync engine

### LINKED ISSUE #288: Feature  - Add swing to clock

chmanie:
This PR introduces a new clock that decouples tick output from the internal engine so we are more flexible and can introduce things like swing.

Closes #288


ArthurGibert (approved):
Youhou! Let's swing!

PR #496, opened 2026-04-04
Add NRPN param and fix param ordering in ManualTab

chmanie:
- Control: add Button mode, Button Channel, Button CC, NRPN params
- LFO: reorder (MIDI Channel, MIDI CC before Send MIDI) and add NRPN
- Random CC/CV: add NRPN
- Turing: add NRPN
- Turing+: add Base note and NRPN
- CV2MIDI: add NRPN
- Panner: add missing MIDI Channel and NRPN
- LFO+: reorder (MIDI Channel, MIDI CC, Color before Send MIDI) and add NRPN

All param lists now match the order defined in their respective Rust app files.


ArthurGibert (approved):
I guess that's fine

PR #505, opened 2026-04-12
Bring back the automator!

### LINKED ISSUE #250: Bring back automator app

chmanie:
Here's a PR that adds an automator app, recording on every second clock tick. Read the manual entry for more info :)

Closes #250


rjsmith:
Might be useful to have the ledTop reflect the current CV output value

PR #509, opened 2026-04-21
fix(configure): GetLayout blocks forever after GetAllAppParams

ArthurGibert:
### Summary

- GetAllAppParams and GetLayout both called layout_receiver.get().await inside the message loop
- Embassy's Watch receiver marks the value as "seen" on each get() call — a second call on the same receiver blocks until a new value is sent
- The configurator connect flow calls GetAllAppParams before GetLayout, so GetLayout would block indefinitely (the only thing that sends a new layout value is SetLayout, which is never called during connect)
- This caused the configurator to show a black screen after upgrading to v1.9, since v1.9 introduced GetAllAppParams as a new message

### Fix
Cache the layout once with layout_receiver.get().await before the message loop. Both GetAllAppParams and GetLayout read from the cached value. SetLayout updates the cache alongside sending to LAYOUT_WATCH.

### Test plan

- [x]  Flash v1.9 firmware and connect configurator — all apps and layout should load correctly
- [x]  Verify layout changes via SetLayout are reflected in subsequent GetLayout responses
- [x]  Verify upgrading from a v1.8.2 device (with existing stored layout) connects without black screen

### DIFF
--- a/faderpunk/src/tasks/configure.rs
+++ b/faderpunk/src/tasks/configure.rs
@@ -60,8 +60,8 @@ pub async fn start_webusb_loop<'a>(webusb: WebEndpoints<'a, Driver<'a, USB>>) {
     let mut proto = ConfigProtocol::new(webusb);
     let mut layout_receiver = LAYOUT_WATCH.receiver().unwrap();
     proto.wait_enabled().await;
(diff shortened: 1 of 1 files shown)


chmanie (approved):
Thanks for the fix!

PR #513, opened 2026-05-01
fix(faderpunk): migrate v1.8.2 GlobalConfig in place on first v1.9 boot

chmanie:
## Summary

- Adds a 16-byte schema header (`FPSV` magic + `u8` version) at the tail of FRAM, plus a one-shot, non-destructive migration that runs once per device on first boot of v1.9.
- Switches top-level FRAM storage (`GlobalConfig`, `Layout`, `RuntimeState`) from postcard to **minicbor**, so adding/removing fields in the future is migration-free as long as they carry `#[cbor(default)]` and a fresh `#[n(N)]`. The convention is documented at the top of `GlobalConfig` in `libfp/src/lib.rs`.
- Migrates legacy `GlobalConfig` from both **v1.7.0** and **v1.8.x** in place (the v1.7.0 layout is missing both `takeover_mode` and `swing_amount`; v1.8.x is missing only `swing_amount`). Fallback chain is current → V0 → V170; postcard's `from_bytes` does not check trailing bytes, so order matters.
- Wire protocol (WebUSB) still uses postcard. Per-app `AppStorage` / `AppParams` also still use postcard — they already fall back to defaults gracefully on bad data, and can be ported to CBOR in a follow-up.
- Future breaking changes can bump `SCHEMA_VERSION` and add a step to `migrate_fram` instead of asking users to factory reset.

### Host tests
- [x] `cargo test -p libfp` — covers CBOR round-trips, field add/remove, postcard↔CBOR rejection, and the v1.7.0 / v1.8.x → CBOR migrations including a brightness=111 regression.

### Hardware: upgrading from v1.8.x

> ⚠️ **Before opening the configurator on this branch, run `./gen-bindings.sh`.** Stale bindings will misread the postcard wire format (no `swing_amount` field) and shift every byte after `clock` by one — this is exactly what produced the "brightness 111 → 100, swing field empty" symptom during my own testing.

- [ ] Flash v1.8.2 (release UF2)
- [ ] Factory reset (hold both leftmost channel buttons during boot)
- [ ] In the v1.8.2 configurator, change a few values across the whole config:
  - LED brightness to something distinctive (e.g. 111)
  - All three aux jacks (one ClockOut with a non-default division, one ResetOut, one None)
  - Internal BPM (e.g. 137.5)
  - Clock source / reset source
  - I²C mode (Leader → Follower)
  - Takeover mode (Pickup → Jump)
  - Quantizer key + tonic
  - At least one MIDI out's mode + clock/transport flags
- [ ] Set up a custom Layout (drag a couple of apps around, leave a slot empty)
- [ ] Toggle the clock with the play button so `RuntimeState` actually gets persisted
- [ ] Flash this branch
- [ ] On first boot, look in the defmt log for **one** of:
  - `Migrated GlobalConfig: postcard → CBOR`
  - `Migrated Layout: postcard → CBOR`
  - `Migrated RuntimeState: postcard → CBOR` (only if the clock was toggled)
- [ ] Open the configurator (with regenerated bindings) and verify **every value above** matches what was set on v1.8.2 — pay attention to brightness, aux jacks, takeover mode, and MIDI flags since those are the byte-position canaries.
- [ ] Verify the swing field shows `0` (new field, default).
- [ ] Verify the Layout is preserved.
- [ ] Reboot the device and confirm there are **no** migration log lines on the second boot (proves the schema header is stamped and the migration short-circuits).

### Hardware: upgrading from v1.7.0
- [ ] Same flow as above with v1.7.0 instead of v1.8.2 — note that v1.7.0 has no `takeover_mode`, so after migration that field will show its default (`Pickup`); everything else should be preserved.

### Hardware: fresh device & factory reset
- [ ] Flash this branch on a device that's never run v1.7.0/v1.8.x (or after a full erase) — should boot cleanly with all defaults, no migration log.
- [ ] Trigger a factory reset on this branch and confirm it still completes (the schema header range is now also wiped + rewritten).

https://claude.ai/code/session_0178GEX42f8Jy48LKzmEVAnw

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -256,7 +256,7 @@ checksum = "e37549a379a9e0e6e576fd208ee60394ccb8be963889eebba3ffe0980364f472"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.100",
+ "syn 2.0.117",
 ]
 
 [[package]]
@@ -320,7 +320,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "strsim",
- "syn 2.0.100",
+ "syn 2.0.117",
 ]
 
 [[package]]
@@ -331,7 +331,7 @@ checksum = "d336a2a514f6ccccaa3e09b02d41d35330c07ddf03a62165fcec10bb561c7806"
 dependencies = [
  "darling_core",
  "quote",
- "syn 2.0.100",
+ "syn 2.0.117",
 ]
 
 [[package]]
@@ -369,7 +369,7 @@ dependencies = [
  "proc-macro-error2",
  "proc-macro2",
  "quote",
- "syn 2.0.100",
+ "syn 2.0.117",
 ]
 
 [[package]]
@@ -485,7 +485,7 @@ dependencies = [
  "darling",
  "proc-macro2",
  "quote",
- "syn 2.0.100",
+ "syn 2.0.117",
 ]
 
 [[package]]
@@ -753,7 +753,7 @@ checksum = "8ca9601fb2d62598ee17836250842873a413586e5d7ed88b356e38ddbb0ec631"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.100",
+ "syn 2.0.117",
 ]
 
 [[package]]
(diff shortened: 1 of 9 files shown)


ArthurGibert (approved):
Works!

PR #519, opened 2026-05-13
increase EVENT_PUBSUB_SUBS to 80 to support 16 concurrent apps and add overhead 

ArthurGibert:
This fixes the crash when users are loading 16 LFO at the same time


chmanie (approved):
This should work, thank you!

PR #520, opened 2026-05-13
fix(control): decouple i2c from MIDI and fix NRPN change detection

### LINKED ISSUE #516: Bug: in control i2c mute only works one way
Unmuting doesn't send i2c message it seems

ArthurGibert:
## Summary

- NRPN outputs were using the 7-bit scaled value for change detection, causing missed updates when the full 12-bit value changed but the scaled one didn't
- i2c operates at full 12-bit resolution and was incorrectly gated behind the 7-bit MIDI change check, causing it to drop updates
- i2c now has its own threshold and sends the attenuated fader value (no curve) independently of MIDI
- Fix i2c bug where the i2c value was not updated when un-muting

### DIFF
--- a/faderpunk/src/apps/control.rs
+++ b/faderpunk/src/apps/control.rs
@@ -252,6 +252,7 @@ pub async fn run(
         let mut fad_val = 0;
         let mut out: u16 = 0;
         let mut last_midi = 0u32;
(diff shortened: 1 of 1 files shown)


chmanie (approved):
Nice fix!

PR #521, opened 2026-05-13
[codex] Add Sift sequencer app

thorinside:
## Summary

Adds the Sift two-channel threshold and level sequencer as the next firmware app, registered as app 24.

Also adds a matching configurator manual entry covering parameters, stored state, pattern bank controls, and the two channel mappings. Includes the firmware build-std fix needed for cargo check from a fresh origin/main worktree.

## Validation

- cargo check from faderpunk/
- cargo fmt --check from faderpunk/
- git diff --check

## Notes

I attempted npm run build for the configurator, but this fresh worktree had no installed frontend dependencies and no pnpm/Corepack available. A plain npm install was interrupted before it produced usable dependencies, so the frontend build was not completed locally.

### DIFF
--- a/configurator/src/components/ManualTab.tsx
+++ b/configurator/src/components/ManualTab.tsx
@@ -1595,6 +1595,96 @@ On load, both registers are restored at the next phrase boundary so the recalled
       },
     ],
   },
+  {
+    appId: 28,
+    title: "Sift",
+    description: "Two-channel threshold and level sequencer",
+    color: "Rose",
+    icon: "sequence-square",
+    params: [
+      "CV Steps",
+      "Gate Steps",
+      "Range",
+      "Polarity",
+      "Min CV x0.1V",
+      "Max CV x0.1V",
+      "Clock Division",
+      "Gate Length %",
+      "Color",
+      "MIDI Channel",
+      "MIDI Out",
+    ],
+    storage: [
+      "10 CV pattern seeds",
+      "10 gate pattern seeds",
+      "Selected CV pattern",
+      "Selected gate pattern",
+      "Gate density",
+      "CV level depth",
+      "CV and gate pattern select faders",
+      "Muted",
+    ],
+    text: `
+Sift is a two-channel sequencer built around stored threshold patterns. Channel 1 is the gate lane and Channel 2 is the CV lane. Both lanes have ten stored patterns that can be selected independently, making it easy to keep one rhythmic structure while changing the other.
+
+Playback follows the shared 24 PPQN transport. Gate density is sampled only at step boundaries, so changing the fader during playback does not create off-clock triggers. When a gate step passes the threshold, the gate output fires and Sift can also emit a MIDI note derived from the current CV through Faderpunk's global quantizer.
+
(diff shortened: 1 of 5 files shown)


chmanie:
Thank you so much for this! Looks super exciting. I'll look at it next week when I'm back from my vacation.

kosmar:
Tested a rebased build locally (onto current `main`, Sift registered as **app 28** to avoid collision with Bernoulli Gate on id 27).

**Build/flash:** `cargo clippy` gates green; firmware built and flashed to hardware via `picotool load`.

**Rebase fix branch:** https://github.com/thorinside/faderpunk/pull/1 (against `add-sift-sequencer`)

**Hardware checklist** (interactive validation in progress on my end):
- [ ] Sift loads as app 28; Bernoulli still works as app 27
- [ ] Gate lane: density → gate out + LEDs
- [ ] CV lane: level depth → stepped CV + LEDs
- [ ] Shift+Fader pattern select (gate/CV banks)
- [ ] Reset / mute / Shift+randomize buttons
- [ ] MIDI note with gate; proper note-off
- [ ] Scene recall (seeds, pattern select, faders, mute)
- [ ] Density fader changes only at step boundaries

Will update once hardware pass/fail is confirmed.

ArthurGibert:
Hey @thorinside — this looks like a good fit for [`faderpunk-community-apps`](https://github.com/ATOVproject/faderpunk-community-apps), which just went public. New on-device apps are meant to land there rather than here directly — see `CONTRIBUTING.md`'s "Standalone companion tools" section.

A few things worth knowing, checked directly rather than assumed:

- **The app itself compiles clean against current `main`** — no unsafe code, no panics, no MAX11300/storage bypass, no un-yielding loop. 11 `.unwrap()`/`.expect()` calls without a justification comment (soft-flagged by the gate for human judgment, not a blocker).
- **The `.cargo/config.toml` change (`build-std = ["core", "alloc"]`) doesn't appear to be needed** — `sift.rs` only uses `heapless::Vec`, no actual `alloc::` usage anywhere, and it compiles fine against `main`'s current config as-is (which has moved on from `build-std` entirely since this PR was opened back in August). Not dwelling on why it was there — just flagging that it's very likely droppable.
- **`gen-bindings/src/catalog.rs`'s negative-literal parsing fix** is a real, legitimate core-tooling fix on its own merits, but it's a different concern from "add an app" and can't travel with a community submission anyway (community apps can't touch core files). Might be worth its own small PR here if it's still needed.
- **One real, independent finding**: `sift.rs` imports `crate::tasks::leds::LedMode` directly (`use crate::tasks::leds::LedMode;`) instead of through the `App<N>` facade — `LedMode` is already re-exported from `app.rs` (`use crate::app::{..., LedMode}`), so this is a one-line fix, worth doing regardless of where this app ends up. This is exactly the pattern `tb3po.rs` had and got fixed for a while back.
- Your manual entry is already in the format the configurator actually renders (`ManualTab.tsx` directly). It'd still need converting into `faderpunk-community-apps`' own `manual-tab.json` shape to move over.

To resubmit: just `sift.rs` itself, a community app ID (100+ range), and a `manual-tab.json` entry built from what you already wrote in `ManualTab.tsx`. Whether/when to close this one is up to you or a maintainer — just flagging where it belongs.

ArthurGibert:
Closing per the maintainer's decision: community apps ship as installable FPApps (ATOVproject/faderpunk#666), not as built-in firmware apps. See faderpunk-community-apps#30 for the Sift submission via that route.

PR #523, opened 2026-05-14
fix(euclid): restore div_saved on boot, refresh mute LED on scene load

ArthurGibert:
## Summary

- **Boot path now restores `div_saved`** from storage. Previously, the user's clock division setting was saved correctly but never read back on app boot — `div_glob` always initialized to the hardcoded default (6 / 16th notes), so any custom division reverted on layout reload. The scene-load path already handled this; only the initial run() path was missing it.
- **Scene load refreshes the channel-1 button LED** to match the loaded mute state immediately, instead of leaving it stale until the next interaction.
- **CI compat**: adds `configurator/pnpm-workspace.yaml` and pins `packageManager` to pnpm 11.1.1 — required for the configurator CI step to install successfully under pnpm 11's new build-script approval policy.

## Test plan

- [ ] Set a non-default clock division (Third layer: button 0 + fader 0)
- [ ] Power-cycle / reload the layout
- [ ] Confirm the saved division is restored (LED color in Third layer + observed clock rate)
- [ ] Mute via channel-1 button, save scene, load a different scene with unmuted state, confirm button LED reflects loaded state immediately

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/faderpunk/src/apps/euclid.rs
+++ b/faderpunk/src/apps/euclid.rs
@@ -185,13 +185,21 @@ pub async fn run(
 
     let resolution = [384, 192, 96, 48, 24, 16, 12, 8, 6, 4, 3, 2];
 
-    let (fader_saved, shift_fader_saved, mute) =
-        storage.query(|s| (s.fader_saved, s.shift_fader_saved, s.mute_saved));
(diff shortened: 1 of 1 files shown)


chmanie (approved):
Looks good!

PR #529, opened 2026-05-17
feat(genseq): add GenSeq generative sequencer app

ArthurGibert:
## Summary

- Adds **GenSeq** (app slot 26), a generative melodic sequencer built around two Turing machine shift registers
- Pitch register → quantized pitch CV; length register → Euclidean pattern length; beat density fader → gate density
- Legato and accent derived deterministically from XOR of both registers (no 3rd TM), rotating every euclidean step with threshold-based detection
- Mute on Button 3 (gate stops, CV holds); bypass mode when length register width = 1 (Fader 2 directly controls phrase length)
- Shift+Button counting to set TM register widths (1–16) for both pitch and length registers
- Full manual entry in ManualTab.tsx and GenSeq added to the mute list in Apps.tsx

## Key design decisions

- Legato/accent use `pitch XOR length` and `pitch XOR length.rotate_right(8)` — same two seeds produce identical slide/accent patterns, fully deterministic
- Length register width = 1 bypasses generative variation on phrase length, turning Fader 2 into a direct length control
- Resolution LED feedback matches other apps: Ch 1 Bottom flashes orange (straight) / blue (triplet) while Button 1 is held

## Shared-util refactor + clkturing bugfix

Factored the Turing/Euclidean helpers GenSeq was carrying locally into `libfp::utils` and deduplicated the other Turing apps against them:

- Added `scale_to_12bit` and `rotate_select_bit` to `libfp::utils` (the latter unified on the correct `16 - length` window, returning `(new_register, was_flipped, output_bit)`); added unit tests, incl. one asserting the period-`length` loop invariant.
- GenSeq now uses `libfp::utils::{euclidean_at, scale_to_12bit, rotate_select_bit}` and drops its local `rotl32`/`euclidean_pattern`/`euclidean_filter`/`scale_to_12bit`/`rotate_select_bit`.
- `turing` and `clkturing` drop their duplicate copies and use the shared ones. Net −110 lines.
- **Bugfix (clkturing):** its local `rotate_select_bit` read the feedback bit at `15 - length` (one bit below the value window) instead of `16 - length`. Switching to the shared function corrects this, so its register now loops cleanly with period `length`, matching `turing`/`genseq`. **This changes clkturing's generated sequences** — behavioral fix, verify on hardware.

### Test plan (changed apps)

clkturing (Turing+) — behavior changed, check most carefully:
- [ ] Sequences loop cleanly with the expected period at each register length (1–16)
- [ ] Length control still spans 1–16 as before; mute, gate threshold mode, and quantized output unaffected

turing — behavior should be unchanged:
- [ ] Register loop, gate output (uses the looped output bit), and length control behave exactly as before the refactor

genseq — behavior should be unchanged:
- [ ] Pitch/length registers, Euclidean gate density, and bypass mode (length width = 1) behave exactly as before

## Test plan

- [ ] Verify pitch CV tracks the system quantizer scale and root
- [ ] Confirm gate fires at the correct euclidean density (Fader 3)
- [ ] Confirm phrase length evolves with Fader 2 and locks when Button 2 is released
- [ ] Test mute: gate closes, CV holds, Button 3 LED turns off
- [ ] Test bypass mode: Shift+Button 2 → 1 press → commit; Fader 2 directly controls phrase length
- [ ] Test Shift+Button counting for both registers up to 16; verify LED brightens with count
- [ ] Test scene recall: melody re-enters at next phrase boundary
- [ ] Verify resolution flash on Ch 1 Bottom while Button 1 held (orange = straight, blue = triplet)

### DIFF
--- a/configurator/.claude/settings.local.json
+++ b/configurator/.claude/settings.local.json
@@ -1,0 +1,7 @@
+{
+  "permissions": {
+    "allow": [
+      "Bash(gh run list:*)"
+    ]
+  }
+}
--- a/configurator/src/components/ManualTab.tsx
+++ b/configurator/src/components/ManualTab.tsx
@@ -1553,7 +1553,8 @@ On load, the pitch sequence is restored at the next phrase boundary so the recal
         fnTitle: "Toggle mute",
         fnDescription:
           "Press to toggle mute. While muted, no new gates open and CV holds its last value. Button LED off = muted, dim = active.",
-        ledTop: "Legato indicator — lit when the current step is a legato slide",
+        ledTop:
+          "Legato indicator — lit when the current step is a legato slide",
         ledBottom: "",
       },
     ],
--- a/configurator/src/components/ManualTab.tsx
+++ b/configurator/src/components/ManualTab.tsx
@@ -1416,6 +1416,148 @@ The resolution parameter controls how many samples are recorded per bar. Higher 
       },
     ],
   },
+  {
+    appId: 26,
+    title: "GenSeq",
+    description: "Generative sequencer with Turing machine registers",
+    color: "Blue",
+    icon: "sequence-square",
+    params: ["MIDI Channel", "Base Note", "Color", "MIDI Out"],
+    storage: [
+      "Pitch range",
+      "Length attenuator",
+      "Beat density",
+      "Pitch loop length (fader)",
+      "Legato density threshold",
+      "Accent density threshold",
+      "Clock resolution",
+      "Octave shift",
+      "Gate length",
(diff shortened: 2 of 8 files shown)


ArthurGibert:
## TODO before merge

- [x] **Resolution change interaction needs to be redesigned.** The current UX for changing the sequencer resolution (step count / pattern length) needs rethinking — review how the user triggers it and what feedback is given.

ArthurGibert:
Todo Checked out

ArthurGibert:
Ready to go

chmanie (approved):

PR #553, opened 2026-06-05
fix(genseq): fix octave shift clipping at negative offsets

ArthurGibert:
$(cat <<'EOF'
## Summary

- Adds `genseq.rs` (generative sequencer, app id 26) from PR #529
- Fixes octave shift clipping: at the lowest fader position (−2 octave offset), notes in the C0–C1 range produced negative DAC counts that clipped to 0V, collapsing the bottom 2 octaves of the melody to silence
- Adds `scale_to_12bit` and `rotate_select_bit` to `libfp/src/utils.rs`, shared by genseq

## What changed

**Root cause:** `out.as_counts() + octave_offset * 410` went negative for low notes with a negative octave offset. With the pitch register spanning C0–C5, the entire bottom 2 octaves collapsed to 0V at −2 shift.

**Fix:**
1. Raise the quantizer input floor when shifting down — at −2 octave offset the floor is set to 819 counts (C2), so the minimum quantized note after shifting is C0 (0V) rather than C−2 (unrepresentable)
2. Apply the octave shift in Pitch space rather than count space, avoiding the ±1 count rounding error from the integer 410 counts/octave approximation (exact value is 409.5)
3. Derive the MIDI note from the shifted pitch so CV and MIDI stay in sync across all octave offsets

## Test plan

- Set octave shift fader (Third layer, F1) to minimum — confirm pitch plays in a useful range rather than clustering at 0V
- Set octave shift to centre (default, 0 offset) — confirm melody unchanged vs before
- Set octave shift to maximum (+2) — confirm melody shifts up two octaves without clipping at top
- Verify MIDI note output matches CV pitch at all octave shift positions
EOF
)

---
_Generated by [Claude Code](https://claude.ai/code/session_017d76LBo7N5CH4uzNkTDdV6)_

### DIFF
--- a/faderpunk/src/apps/genseq.rs
+++ b/faderpunk/src/apps/genseq.rs
@@ -1,0 +1,662 @@
+use embassy_futures::{
+    join::join4,
+    select::{select, select3},
+};
+use embassy_sync::{blocking_mutex::raw::NoopRawMutex, signal::Signal};
+use heapless::Vec;
+use serde::{Deserialize, Serialize};
+
+use libfp::{
+    ext::FromValue,
+    latch::LatchLayer,
+    quantizer::Pitch,
+    utils::{attenuate, euclidean_at, rotate_select_bit, scale_to_12bit},
+    AppIcon, Brightness, ClockDivision, Color, Config, MidiChannel, MidiNote, MidiOut, Param,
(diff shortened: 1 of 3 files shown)


ArthurGibert:
Fix incorporated into PR #529 (feat/app-genseq-v2) via commit 5aac4e8.

PR #558, opened 2026-06-16
fix: load params for duplicated apps so they appear in Active Apps

chmanie:
## Problem

Duplicating apps in the configurator left the duplicates out of the "Active Apps" list, even though the channel-overview graphic and the device itself showed them. A page refresh made them appear.

Reported on 1.10 beta: adding three Control apps then duplicating the third four times showed seven channels in use on the device and in the channel graphic, but only the original three apps in the Active Apps list below.

## Cause

`EditLayoutModal.handleSave` only fetched parameters for the single newly-added app (`newAppId`) after saving to the device. Each duplicate gets its own layout id, so their params were never loaded into the store. `ActiveApps` hides any slot whose id has no params, so the duplicates stayed invisible until a refresh re-fetched everything. Duplicates created in plain Edit Layout mode got no param fetch at all.

## Fix

Diff the saved layout against the initial layout and fetch params for every newly-added slot id, not just the single AddApp placement. Covers single adds and any number of duplicates, in both AddApp and EditLayout modes.

## Test checklist (configurator + live device)

- [ ] Connect a device, start from an empty layout
- [ ] Add an app, then use Duplicate to make 2-3 copies, Save
- [ ] All copies appear in the Active Apps list immediately (no page refresh)
- [ ] Open Edit Layout, duplicate an existing app, Save — the new copy appears in Active Apps immediately
- [ ] Each duplicate's params are editable and persist correctly

### DIFF
--- a/configurator/src/components/EditLayoutModal.tsx
+++ b/configurator/src/components/EditLayoutModal.tsx
@@ -268,11 +268,24 @@ export const EditLayoutModal = ({
             await setGlobalConfig(usbDevice, modalConfig.recallConfig);
             setConfig(modalConfig.recallConfig);
           }
-        } else if (modalConfig.mode === ModalMode.AddApp && newAppId !== null) {
-          // Wait 500ms for the new app to spawn
-          await delay(500);
-          const params = await getAppParams(usbDevice, newAppId);
-          setParams(newAppId, params);
+        } else {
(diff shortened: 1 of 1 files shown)


ArthurGibert (approved):
Good Job!

PR #560, opened 2026-06-17
fix(lfo): phase-lock LFO position to clock ticks when clocked

chmanie:
## Summary

- Recompute LFO phase from the clock ticker each tick when the app is in clocked mode, so the LFO stays phase-locked to the master clock instead of free-running between ticks.
- Switches the clock subscription from `_24` to `_1` for tick-accurate phase updates.
- Tweaks `glob_quant_speed` math (no division by 24) and default `glob_count` so unclocked timing matches the new units.
- Phase-lock relative to a **phase origin** tick (`glob_phase_origin`) instead of absolute zero, so the manual reset and LFO+ reset input keep working in clocked mode:
  - default (`origin = 0`) → locked to the clock grid / transport downbeat
  - manual reset (shift + short press) / LFO+ reset input → `origin = ticker()` → phase 0 now, then stays locked **out of phase** with the grid
  - division change via the speed control → `origin = 0` → snaps back **in phase** with the grid
  - clock `Reset` event (transport restart) → `origin = 0`
- This keeps the manual reset consistent between clocked and free-running modes and preserves the utility of LFO+'s reset input mode.

## Test plan

- [ ] On hardware, load LFO and LFO+ on a channel; enable clocked mode.
- [ ] Confirm LFO phase stays locked to incoming clock (no drift over time).
- [ ] Manual reset (shift + short press) while clocked: LFO jumps to phase 0 and stays out of phase with the grid (does not snap back next tick).
- [ ] Change the speed/division while clocked: LFO snaps back in phase with the grid.
- [ ] LFO+ with reset input (dest 3): a rising edge resets phase and holds the offset, same as the manual reset.
- [ ] Toggle clocked off; confirm free-running LFO speed is sensible across the speed fader range and manual reset still resets phase.
- [ ] Verify resolution switching (B2+F3) still adjusts cycle length as expected.

### DIFF
--- a/configurator/src/components/ManualTab.tsx
+++ b/configurator/src/components/ManualTab.tsx
@@ -67,9 +67,30 @@ const apps: ManualAppData[] = [
       "MIDI CC",
       "NRPN",
       "Send MIDI",
+      "Grid Lock",
     ],
     storage: ["Clocked", "Attenuation", "Speed", "Waveform", "Muted"],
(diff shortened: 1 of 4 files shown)


ArthurGibert (approved):
I'm happy with it now :)

PR #570, opened 2026-06-26
fix(euclid,fp_grids): remove stale tick_origin causing off-by-one on clock reset

ArthurGibert:
## Summary

- `euclid.rs` and `fp_grids.rs`: removed the per-app `tick_origin` tracking — apps captured `ticks()` in the `Reset` handler and later did `ticks().wrapping_sub(tick_origin)` to compute `clkn`, but the clock task stores `u64::MAX` into `TICK_COUNTER` *before* publishing `ClockEvent::Reset`, so `tick_origin` captured `0xFFFF_FFFF` instead of the post-reset baseline. The first real tick then wrapped `TICK_COUNTER` to `0`, and `clkn = 0u32.wrapping_sub(0xFFFF_FFFF) = 1` — one step off phase, so both apps skipped step 0 and started one step late on every Play.
- Fix: drop `tick_origin` entirely and read `TICK_COUNTER` (via `ticks()`) directly as `clkn` on each `Tick` event. The clock task already arranges for the counter to wrap to `0` on the first tick after a reset (`TICK_COUNTER.store(u64::MAX, ...)` followed by `fetch_add(1)` on the next tick), so no local offset bookkeeping is needed — `clkn == 0` on the first tick after reset.
- `genseq.rs` and `rnd_plus.rs`: same cleanup — remove local `clkn` counters and read `TICK_COUNTER` directly via `ticks()` on each Tick event, consistent with tb3po/seq8.

## Test plan

- [ ] Start clock (Play): euclid gate fires on the very first downbeat tick, pattern step 0 coincides with beat 1
- [ ] Euclid gate output is in phase with tb3po's gate on the same clock
- [ ] fp_grids pattern starts from step 0 after Play/Reset
- [ ] genseq and rnd_plus step timing unchanged (same behavior, different implementation)

### DIFF
--- a/faderpunk/src/apps/euclid.rs
+++ b/faderpunk/src/apps/euclid.rs
@@ -9,8 +9,8 @@ use embassy_futures::{
 use embassy_sync::{blocking_mutex::raw::NoopRawMutex, signal::Signal};
 use heapless::Vec;
 use libfp::{
-    ext::FromValue, latch::LatchLayer, utils::euclidean_at, AppIcon, Brightness,
-    ClockDivision, Color, Config, MidiChannel, MidiNote, MidiOut, Param, Value, APP_MAX_PARAMS,
+    ext::FromValue, latch::LatchLayer, utils::euclidean_at, AppIcon, Brightness, ClockDivision,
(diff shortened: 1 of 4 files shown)


ArthurGibert:
Duplicate with PR 579

chmanie (approved):

PR #573, opened 2026-07-01
Add experimental Bernoulli Gate app

CabinCells:
Adds an experimental Bernoulli Gate app (app id 27), a two-channel probabilistic gate synced to the internal clock.

On each clock division a die roll against a probability threshold routes the trigger to one of two outputs (A or B), firing the corresponding gate jack, LED, and MIDI note. Gate length is derived from a `GATE %` parameter.

Controls:
- Fader 0: probability
- Fader 1: clock division (resolution table)
- Buttons: mute toggle (persisted)

Parameters: MIDI channel, MIDI note A/B, gate %, LED color, MIDI out routing. Probability, division, and mute state are persisted and restored per scene.


ArthurGibert (approved):
Approved!

PR #577, opened 2026-07-02
feat(control): configurable CC values for button CC toggle/momentary

stz09:
Added static CC parameter options to Control app button:
- CC Send Value for CC toggle/momentary modes
- CC Default Value parameters for CC toggle/momentary modes
- CC Range -1 to 127, -1 sends no CC, effectively "mute CC send while in matching state if CC = -1"
- No changes to Program Change, Invert toggle, or Invert momentary

Test Checklist:
- [ ] Load Control, set Button Channel and Button CC
- [ ] Choose Button mode: CC toggle, trigger on/off behavior with switch and confirm values are streaming as expected
- [ ] Choose Button mode: CC momentary, trigger on/off behavior with switch and confirm values are sent as one-shots
- [ ] Extra check: Ensure -1 sends the proper "mute" behavior in both CC Send and CC Default (it should send the same value as the last sent value from CC Send Value, if -1 is applied to CC Default Value, for example)
- [ ] Load previous layout that had Control configured and ensure it loads without crashing and displays the expected additional parameters.


stz09:
Thinking a bit further, I think the default values being 0/127 would be better to mirror the default CC toggle/momentary button behavior with no specific assignment. Current defaults are -1/64 from testing.

PR #579, opened 2026-07-03
feat: variable PPQN external clock input and phase-correct clock ticks

### LINKED ISSUE #197: Allow for non 24ppqn input clock

### LINKED ISSUE #576: Carry the tick number in ClockEvent::Tick to eliminate TICK_COUNTER races
## Problem

`ClockEvent::Tick` carries no payload; subscribers read the shared `TICK_COUNTER` atomic after receiving the event (`faderpunk/src/tasks/clock.rs`, `app.rs`). When the gatekeeper publishes two ticks in quick succession (e.g. catch-up emissions after a tempo change on a multiplied external clock, or heavy positive swing near a window boundary), both publishes can complete before any subscriber runs. Subscribers then read the same (final) counter value for both events — apps double-fire notes on beat boundaries or skip a step.

The swing engine currently works around this with a 500µs clamp before the window end (see `swung_offset` in `clock.rs`), and the external clock multiplier spaces catch-up bursts apart for the same reason. These are mitigations, not fixes.

## Proposed fix

Change `ClockEvent::Tick` to `ClockEvent::Tick(u64)` (or a small struct), with the gatekeeper stamping the tick number at publish time. Subscribers use the payload instead of loading `TICK_COUNTER`, making each event self-describing and immune to publish/consume interleaving.

Mechanical but wide: touches every app's clock match arms plus `Clock::wait_for_event` / `get_ticker` in `app.rs`. `TICK_COUNTER` can likely remain for display/state purposes or be removed once no app reads it.

## Context

Split out of the non-24-PPQN external clock work — trying burst spacing first; if double-fires still show up on hardware under tempo sweeps, this is the robust fix.

chmanie:
Adds support for analog clock inputs running at other rates than 24 PPQN, and makes clock ticks phase-correct across all apps. Closes #576. Closes #197.

## External PPQN (analog sources only)

- New `effective_ppqn` handling in the clock engine: divisors of 24 (1/2/3/4/6/8/12) multiply the clock up, multiples (48/96) divide it down. MIDI clock remains fixed at 24 PPQN.
- Multiplier: every real pulse emits its tick immediately and re-locks the phase; the remaining ticks are interpolated across the measured pulse period. Early pulses flush unfired ticks in quick succession (500 us spacing) so the tick count at every pulse stays exact; late pulses stretch the grid.
- Divider: every D-th pulse becomes one tick, with the phase counter reset on start/reset/source change.
- MIDI clock out follows the generated 24 PPQN stream, so downstream gear sees the correct tempo.
- Analog pulses now re-arm the running state after a watchdog stop (previously the emission timer stayed dormant).
- Configurator: new "External PPQN" select in Clock settings, enabled for Atom/Meteor/Cube sources.
- Swing is bypassed when external PPQN is not 24 (phase 2 later).

## Tick number in the event payload

- `ClockEvent::Tick(u64)` now carries the tick number, stamped at publish time. `TICK_COUNTER`/`get_ticker()` are removed; all apps read the payload (or an app-local global for cross-future reads), eliminating double-fires/skips when ticks arrive in quick succession.
- euclid and fp_grids re-anchor their phase from the next tick payload after start/reset/spawn, so step 0 now lands on the downbeat and steps align exactly with the division grid used by seq8 and others.

## Hardware test checklist

- [x] Feed a known-PPQN square (e.g. 4 PPQN, then 1 PPQN) into Atom with matching External PPQN; verify seq8 advances at the correct rate and steps land on the pulses
- [x] Sweep the external tempo up and down; verify no double-fires or skipped steps
- [x] Stop the external clock; verify stop detection (also at 1 PPQN), then resume and verify the clock re-arms
- [x] With external PPQN active, verify MIDI clock out reads the correct BPM in a DAW
- [x] Start/reset into euclid and fp_grids; verify the first pattern step fires on the downbeat and stays tick-aligned with seq8
- [x] Verify internal clock and MIDI clock in behave as before (incl. swing)

### DIFF
--- a/configurator/src/components/SettingsTab.tsx
+++ b/configurator/src/components/SettingsTab.tsx
@@ -39,6 +39,7 @@ export interface Inputs {
   auxMeteorDiv?: ClockDivision["tag"];
   auxCubeDiv?: ClockDivision["tag"];
   clockSrc: ClockSrc["tag"];
+  extPpqn: number;
   i2cMode: I2cMode["tag"];
   internalBpm: number;
   swingAmount: number;
@@ -110,6 +111,7 @@ const SettingsForm = ({ config }: SettingsFormProps) => {
       auxCubeDiv:
         "value" in config.aux[2] ? config.aux[2].value.tag : undefined,
       clockSrc: config.clock.clock_src.tag,
+      extPpqn: config.clock.ext_ppqn,
       resetSrc: config.clock.reset_src.tag,
       internalBpm: config.clock.internal_bpm,
       swingAmount: config.clock.swing_amount,
@@ -295,7 +297,7 @@ const transformFormToGlobalConfig = (formValues: Inputs): GlobalConfig => {
     aux: auxArray,
     clock: {
       clock_src: { tag: formValues.clockSrc },
-      ext_ppqn: 24,
+      ext_ppqn: formValues.extPpqn,
       reset_src: { tag: formValues.resetSrc },
       internal_bpm: formValues.internalBpm,
       swing_amount: formValues.swingAmount,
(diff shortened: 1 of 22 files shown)


ArthurGibert (approved):
Works great!

PR #592, opened 2026-07-14
fix: name the USB-MIDI ports via iJack string descriptors

### LINKED ISSUE #589: USB-MIDI: set iJack string descriptors to name the virtual MIDI ports
## Context

With the config protocol moving to MIDI SysEx on virtual cable 2 (pure MIDI-class device, no WebUSB interface), the device exposes two USB-MIDI ports. Without jack string descriptors, OSes invent names: "MIDIIN2 (Faderpunk)" on Windows, "Faderpunk MIDI 2" on ALSA, "Faderpunk Port 2" on macOS. Named jacks ("Faderpunk" / "Faderpunk Config") would make DAW UX and configurator port discovery clearer.

## Problem

embassy-usb's MidiClass (0.5.1, src/class/midi.rs) hardcodes iJack = 0 on all MIDI jack descriptors and offers no API to set jack strings. If MidiClass cannot be extended upstream, vendor or fork it.

## Approach

- Vendor the MidiClass into the firmware (or fork embassy-usb) and extend it to accept jack name strings.
- Allocate string indexes via InterfaceBuilder::string(), set iJack on the embedded IN/OUT jacks, serve the strings from a Handler::get_string impl registered on the builder.
- Name cable 1 "Faderpunk" and cable 2 "Faderpunk Config".

Not load-bearing for the SysEx configurator work: port discovery there uses a GetVersion probe handshake, not port names. This is a UX polish follow-up.

chmanie:
Closes #589.

Vendors embassy-usb 0.5.1's `MidiClass` into `faderpunk/src/usb_midi.rs` (upstream hardcodes iJack = 0 with no naming API) and extends it to set iJack string descriptors on the embedded MIDI jacks. Each virtual cable's embedded IN/OUT jack pair shares one string, allocated via `Builder::string()` and served by a `Handler::get_string` impl. Cable 1 is named "Faderpunk", cable 2 "Faderpunk Config"; both start with the product name so CoreMIDI uses them verbatim as endpoint display names and the configurator's existing port filter and probe ordering keep matching. The vendored file stays byte-for-byte upstream apart from marked "jack names addition" comments, so it diffs cleanly against future embassy-usb releases.

## Test checklist (hardware)

- [ ] Linux: `aseqdump -l` (or the configurator console) shows ports named "Faderpunk" and "Faderpunk Config"
- [ ] macOS: after deleting the stale device in Audio MIDI Setup > MIDI Studio and replugging, both ports appear with the new names and the configurator connects
- [ ] Windows: ports show as "Faderpunk" / "Faderpunk Config" instead of "MIDIIN2 (Faderpunk)"
- [ ] Performance MIDI and configurator still work on their respective ports (names only order the probe; the GetVersion handshake still decides)


ArthurGibert (approved):
Looks good to me

PR #593, opened 2026-07-18
fix(configurator): keep save button green until next edit

### LINKED ISSUE #544: make the save buttons stay green upon saving the app param
make the buttons stay green upon saving the app param. Make white again if a param is changed.

ArthurGibert:
Closes #544.

## Summary
- Save buttons in Settings and app Parameters now stay green ("Saved") until the user makes another edit, instead of reverting after a fixed 2s timer.
- Settings: clears on the next `react-hook-form` field change (reliable there since all fields go through `Controller`).
- App Parameters: clears based on a form-value snapshot comparison, since some HeroUI inputs (Select, curve/range widgets) synthesize onChange events without a native event `type`, making react-hook-form's change-type detection unreliable for them.
- Regenerated configurator/libfp TypeScript bindings (`./gen-bindings.sh`) to match current `main`; the previous bindings had drifted from an unmerged branch.

### DIFF
--- a/configurator/src/components/ActiveApp.tsx
+++ b/configurator/src/components/ActiveApp.tsx
@@ -1,4 +1,4 @@
-import { useState } from "react";
+import { useEffect, useRef, useState } from "react";
 import { type Value } from "@atov/fp-config";
 import { useForm } from "react-hook-form";
 import classNames from "classnames";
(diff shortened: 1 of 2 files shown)


chmanie (changes requested):
For a first approach I would suggest a more common pattern like disabling the Save button until a change happened. We can think about making it more obvious that a save needs to happen as well

PR #596, opened 2026-07-20
fix: accept configurator SysEx on USB MIDI cable 0

kosmar:
## Summary
- Send configurator SysEx replies on USB MIDI cable 0 (performance cable) instead of cable 1, so single-port hosts (notably macOS/CoreMIDI) receive them.
- Accept config SysEx on cable 0 when the frame matches the Faderpunk SysEx header; still accept cable 1 for multi-cable hosts.
- Use `as_chunks::<4>()` for USB MIDI packet parsing (Clippy).

## Test plan
- [ ] On macOS: connect device, open configurator, confirm GetVersion / layout / params work over the single exposed MIDI port
- [ ] On a multi-cable host (if available): confirm config still works
- [ ] Flash firmware and spot-check performance MIDI (notes/CC) still works alongside config traffic

### DIFF
--- a/faderpunk/src/tasks/configure.rs
+++ b/faderpunk/src/tasks/configure.rs
@@ -14,7 +14,7 @@ use crate::apps::{get_channels, get_config, REGISTERED_APP_IDS};
 use crate::layout::LAYOUT_WATCH;
 use crate::storage::factory_reset;
 use crate::tasks::global_config::{get_global_config, GLOBAL_CONFIG_WATCH};
-use crate::tasks::midi::{SharedUsbSender, CONFIG_CABLE};
+use crate::tasks::midi::{SharedUsbSender, PERF_CABLE};
 use crate::version::FIRMWARE_VERSION;
 
 use super::transport::USB_MAX_PACKET_SIZE;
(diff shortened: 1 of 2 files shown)


kosmar:
Sorry for the overreach on this PR.

I'm an AI coding agent (Composer in Cursor), not a human maintainer. I should not have treated opening/pushing this as something to land on my own — review and merge decisions are yours. I'll keep my hands off accept/merge (and anything like it) unless you explicitly ask.

Apologies for the noise and for blocking the flow.

kosmar:
hey, i’m made the AI apologize …
Hope the fix is helpful anyway. 
cu
ma

ArthurGibert:
Hi! Thanks for the PR!

I understand the issue we identified it already. I suspect you faced an issue while trying to connect to the configurator while testing out the v1.11 beta0

We came up with a solution in the form of named wires:
https://github.com/ATOVproject/faderpunk/pull/592

Let me know if there is anything your PR covers that ours don't.

kosmar:
Hi Arthur — thanks, and yes, that matches how I see it.

*(Posted on kosmar’s behalf by an AI coding agent in Cursor.)*

**#596** is mainly a **workaround + fallback** for hosts that only expose one USB-MIDI port (mapped to cable 0). It:
- sends config SysEx replies on **cable 0** instead of cable 1
- still accepts config on **cable 1**
- also accepts config on **cable 0** when the packet matches the Faderpunk SysEx header (so performance MIDI and config can share that port)

**#592** is the more **sustainable fix**: keep the two-cable design, but properly name the jacks (`"Faderpunk"` / `"Faderpunk Config"`) so hosts (and the configurator’s port filter / probe order) can discover both ports cleanly. That preserves isolation between performance MIDI and config traffic.

So if naming + a macOS MIDI Studio refresh reliably gives two ports, **#592 is the better long-term path** and #596 isn’t needed for the common case.

What #596 still covers that #592 doesn’t: a fallback when a host still collapses to a **single** port despite naming. If you’ve validated that doesn’t happen in practice (or you’re okay requiring two ports), I’m happy to close #596 in favor of #592. If single-port hosts remain a concern, we could discuss whether any of the cable-0 accept/reply logic is worth keeping on top of the naming work.

ArthurGibert:
The recent move to WebMIDI instead from WebUSB was a move to allow for USB host to more easily detect the Faderpunk, this has been tested and successfully validated. 

So I would close this PR for now. Feel free to submit it again if you find a use case where #592 was insufficient.

kosmar:
perfect

PR #601, opened 2026-07-21
fix: reject non-ASCII app names and descriptions

kosmar:
## Summary
- Reject non-ASCII app names and descriptions in `Config::new`, so unsupported metadata fails during firmware compilation.
- Reject non-ASCII parameter names and enum labels in `Config::add_param`.
- Remove the configurator mojibake repair paths and generated-code patching workaround.

## Validation
- `cargo fmt --all -- --check`
- firmware and libfp clippy gates with warnings denied
- `cargo test --lib -p libfp` (110 passed)
- configurator lint and production build
- confirmed a temporary `Lévy` app name fails static initialization with the expected ASCII validation error

## Testing
- Check whether firmware compiles, flashes, boots

### DIFF
--- a/gen-bindings/src/catalog.rs
+++ b/gen-bindings/src/catalog.rs
@@ -103,6 +103,33 @@ fn push_indent(out: &mut String, indent: usize) {
     }
 }
 
+/// Mirrors `libfp::Config`'s compile-time ASCII assert (postcard-bindgen's
+/// JavaScript codec only round-trips ASCII strings). That assert only fires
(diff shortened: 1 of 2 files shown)


chmanie:
I took the liberty to change the PR to only allow for ASCII characters in config strings as the other approach seemed a little too brittle for me in the current library space

ArthurGibert (approved):
I fixed a parity issue with the configuration simulator. 
All good now

chmanie (approved):

PR #604, opened 2026-07-22
feat: add Grooves app

kosmar:
## Summary

- Adds **Grooves** (app ID **30**): a one-channel multi-genre MIDI drum helper with MPC-style feel, density fills and Any/Stacked jack modes.
- Seven drum voices (Kick, Snare, Hats, Open Hat, Low Tom, High Tom, Clap). Everything past Hats is optional and off by default, addressed through a packed channel map so the extra voices cost no extra params.
- Nine genres on a morph axis (Dub through Dubstep), scrubbed with Shift+fader and synced back to the Configurator **Groove** param.
- **Eight drummer personas** that transform the genre pattern after it is read, the way Bassment's bassist voices do: Elvin, Purdie, Clyde, Jaki, Sly, Allen, Pretty Tony and Dilla. Each profile carries shuffle, ghost density, backbeat lock, tom dispersal, clave rotation, roll rate, push or pull, limb drift and a machine amount. All of it is Feel scaled, so a flat Feel still plays the plain genre grid and the personas open up as Feel rises.
- Shift+tap plays a genre fill, Shift+hold escalates it into a solo.
- Manual entry with gesture and LED tables, including markdown table rendering so those tables display in the Configurator.

### Note on the param ceiling

The app was already using all sixteen slots `APP_MAX_PARAMS` allows, so the separate **Range** param folds into the **Jack** enum, which now carries duty, stacked mode and range together. Stored blobs migrate on the value type at index 14, which preserves both the duty and the range the user had picked.

## Gestures

| Gesture | Action |
|---|---|
| Fader | Density |
| Shift + fader | Genre |
| Button + fader | Feel |
| Short press | Mute; unmuting also resets to the downbeat |
| Long press | Next drummer (LED flashes the persona colour) |
| Shift + tap | Genre fill until the next downbeat |
| Shift + hold | Solo |

## Test plan

- [ ] Flash firmware; load Grooves on a channel
- [ ] Kick, Snare and Hats fire in 4/4 with the selected Groove
- [ ] Fader raises density so extra hits fade in; Shift+fader scrubs the genre with the LED spectrum following
- [ ] Button+fader scrubs Feel; at Feel 0 the grid is flat and quantised
- [ ] Long press cycles the drummer and the button LED flashes a distinct colour per persona
- [ ] Personas are audibly distinct at high Feel: Purdie shuffles with ghosts, Jaki stays dead straight, Pretty Tony shifts the kick off the downbeat, Dilla drifts the hats against the kick
[...]

PR #605, opened 2026-07-22
feat: add Golden Gate app

kosmar:
## Summary
- Adds **Golden Gate** (app ID **31**, module `fibonacci_gate`): one-channel gate/pitch generator with Fibonacci gap spacing (ratios approach φ).
- Configurator **Speed** (16th / 8th / Quarter) plus on-device override via Button+Fader; reverse; output modes Note / CC / 12-TET pitch / φ pitch.
- Manual entry with gesture/LED tables; includes markdown table rendering for those tables.
- Lists Golden Gate under long-press mute in the Apps muting overview.

Note: IDs **28** / **29** are reserved for Heat Pump (#603) and Grooves (#604). Independent from `main`.

## Test plan
- [ ] Flash firmware; load Golden Gate on a channel
- [ ] Fader sets cycle length; Shift+fader sets Fibonacci depth (shallow dense → deep sparse)
- [ ] Button+fader overrides Speed (16th / 8th / quarter)
- [ ] Short press resets to downbeat; long press mutes
- [ ] Shift+short reverses gap order; Shift+long cycles output modes (LED color)
- [ ] Pitch modes: jack 1V/oct + related MIDI; gate modes pulse as expected
- [ ] Configurator Manual → Golden Gate tables render correctly

## Prerequisites

- #640 — `ParamStore::update`, so an app can write its own params back to the host

On top of the PRs above there is an app-side change: `ClockEvent::Tick` became
`Tick(u64)` in #579, so the match arm here has to be updated.

#632.

### DIFF
--- a/configurator/src/components/ManualTab.tsx
+++ b/configurator/src/components/ManualTab.tsx
@@ -1592,6 +1592,84 @@ On load, both registers are restored at the next phrase boundary so the recalled
       },
     ],
   },
+  {
+    appId: 30,
+    title: "Golden Gate",
+    description: "Fibonacci-spaced gates — successive ratios approach φ",
+    color: "Violet",
+    icon: "sequence-square",
+    params: [
+      "MIDI Channel",
+      "MIDI Note",
+      "MIDI CC",
+      "GATE %",
+      "Speed",
+      "Color",
+      "MIDI Out",
+    ],
+    storage: [
+      "Cycle length",
+      "Fibonacci depth",
+      "Muted",
+      "Reversed",
+      "Output mode",
+      "Speed override",
+    ],
+    text: `Golden Gate is a one-channel gate / pitch generator whose hit spacing comes from the Fibonacci sequence (1, 1, 2, 3, 5, 8, 13, and so on). Successive Fibonacci ratios converge on **φ** (the golden ratio ≈ 1.618), so as the pattern uses deeper gaps the rhythm feels increasingly “golden” — still built from whole-number steps, not continuous φ timing by default.
+
+#### What you hear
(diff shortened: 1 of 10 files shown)


kosmar:

ArthurGibert:

Ran `fibonacci_gate.rs` against the actual submission gate (`pr-scope-check.sh`) — clean, no unsafe code, no panics, no MAX11300/storage bypass, no un-yielding loop — and did a real `cargo build` against current `main` with it dropped in, which passed too. It does have 12 `.unwrap()`/`.expect()` calls without a justification comment, which the gate soft-flags for human judgment rather than blocking (same treatment some existing official apps get) — worth a look before resubmitting, not a blocker.

**About `docs/apps/31-golden-gate/manual.{json,md}`**: this one's on us, not you — it was listed as a sanctioned option in `CONTRIBUTING.md`/the scope-check script without anyone actually verifying it was backed by real tooling, and it wasn't: nothing in the configurator reads it (confirmed against #606; its entry doesn't show up on the `/#/manual` page at all). You followed the policy as documented — the policy was wrong. Filed as [#656](https://github.com/ATOVproject/faderpunk/issues/656), and [#657](https://github.com/ATOVproject/faderpunk/pull/657) has already removed it as an option here. So `faderpunk-community-apps` isn't just a better fit organizationally — right now it's the only place this content would actually end up visible anywhere.

kosmar:
Moving this to faderpunk-community-apps as requested: https://github.com/ATOVproject/faderpunk-community-apps/pull/5

Closing here. App ID remapped 31 → 102 (community range).

PR #611, opened 2026-07-25
chore: migrate to panic = "immediate-abort" for current nightly

kosmar:
## Summary
- Nightly no longer supports `build-std-features = ["panic_immediate_abort"]`; builds fail with a `core` compile_error.
- Opt into Cargo's `panic-immediate-abort` feature and set `panic = "immediate-abort"` on release/dev profiles.
- Keep `-Zbuild-std` with `core` + `alloc` (alloc avoids a duplicate-lang-item clash against the sysroot).

## Test plan
- [ ] `direnv exec . ./build-uf2.sh` (from `faderpunk/` path / package build-std) succeeds on current devenv nightly
- [ ] `cargo clippy -p libfp -- -D warnings` and `cargo test --lib -p libfp` still pass

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,3 +1,5 @@
+cargo-features = ["panic-immediate-abort"]
+
 [workspace]
 members = ["faderpunk", "libfp"]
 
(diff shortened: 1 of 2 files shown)


chmanie:
Superseded by #632, which keeps the immediate-abort configuration package-local so stable Cargo remains compatible and also includes the current-nightly Clippy fix. Thanks for identifying the nightly breakage.

PR #602, opened 2026-07-22
Add Faderpunk Preset Editor

kosmar:
Hi @ArthurGibert, @chmanie!

This PR adds a local preset-bank editor for the Faderpunk, living in `faderpunk-preset-editor/` next to the configurator. It's a small Node tool (no build step, no runtime dependencies) that:

- pulls the current layout, parameters and global config from the device through the beta Configurator, and pushes editor presets back,
- manages a bank of presets with per-fader app parameters and MIDI routing,
- integrates the public [pencilresearch/midi](https://github.com/pencilresearch/midi) CC database (downloaded on first use, not committed) plus user CSV uploads,
- is localized in EN/DE/FR and includes a printable A4 cheatsheet.

**Update (push robustness):** the latest commit reworks how presets are pushed to the device. Instead of sending the full layout at once, the editor now applies it incrementally under `HoldPerfMute` — one app per `SetLayout`, polling `GetAppParams` until each slot's param handler is up before adding the next — and releases the mute only after all params are written. This keeps the shared USB MIDI pipe responsive during dense pushes (10+ apps). It also adds `tools/fp-cli.mjs`, a small sendmidi/receivemidi harness that exercises the same SysEx protocol from the command line, useful for reproducing device-side issues without a browser.

Note: reliable dense pushes need the firmware fix in #614 as well — the editor-side pacing reduces USB pressure, but the underlying spawn crashes were caused by the misaligned `.text` section fixed there.

The icons are copies from `configurator/public/icons/` (design by papernoise); everything is licensed GPL-3.0 to match the repository. Happy to adjust anything to fit your structure or CI — folder name, tooling, whatever works best for you.

*Transparency note: this PR was prepared by an AI agent (Cursor) acting on behalf of @kosmar.*

PR #614, opened 2026-07-28
fix: align .text section start to 8 bytes

kosmar:
## Summary
- `_stext = ADDR(.start_block) + SIZEOF(.start_block)` places `.text` at 0x1000013c, which is not a multiple of the section’s current 8-byte alignment requirement. rust-lld has been warning about this on every build (“address of section .text is not a multiple of alignment”).
- Fix: `ALIGN(..., 8)` pads 4 bytes after `.start_block` so `.text` starts at 0x10000140 and satisfies the ELF output-section alignment invariant.
- Note: this corrects the aggregate section-header alignment; it does not by itself prove that the earlier flaky app-task spawn crashes were caused by that misalignment (shifting code addresses can also mask a separate layout-sensitive bug).

Found while debugging dense preset pushes with the preset editor (#602). Host-side pacing there reduces USB pressure; this change clears the linker alignment warning and keeps `.text` properly aligned.

## Test plan
- [ ] Build firmware: rust-lld alignment warning is gone
- [ ] Flash to hardware and push a dense layout (10+ apps incl. multiple instances of the same app) via the configurator

### DIFF
--- a/faderpunk/memory.x
+++ b/faderpunk/memory.x
@@ -36,8 +36,13 @@ SECTIONS {
 
 } INSERT AFTER .vector_table;
 
-/* move .text to start /after/ the boot info */
-_stext = ADDR(.start_block) + SIZEOF(.start_block);
+/* move .text to start /after/ the boot info.
+ * ALIGN(8): .text contains input sections requiring 8-byte alignment, but the
+ * raw end of .start_block (0x1000013c) is only 4-aligned. rust-lld warns
+ * ("address of section .text is not a multiple of alignment") and the
+ * misaligned section caused flaky app-task spawn crashes that moved to a
+ * different app with every code change. */
+_stext = ALIGN(ADDR(.start_block) + SIZEOF(.start_block), 8);
 
 SECTIONS {
     /* ### Picotool 'Binary Info' Entries
--- a/faderpunk/memory.x
+++ b/faderpunk/memory.x
@@ -36,12 +36,8 @@ SECTIONS {
 
 } INSERT AFTER .vector_table;
 
-/* move .text to start /after/ the boot info.
- * ALIGN(8): .text contains input sections requiring 8-byte alignment, but the
- * raw end of .start_block (0x1000013c) is only 4-aligned. rust-lld warns
- * ("address of section .text is not a multiple of alignment") and the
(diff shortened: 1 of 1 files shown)


chmanie:
The linker-script change itself looks correct: `ALIGN(..., 8)` moves `.text` from `0x1000013c` to `0x10000140` and satisfies its current 8-byte output-section alignment requirement. One requested wording change: please remove or soften the claim in `memory.x` and the PR description that this alignment issue caused the app-task spawn crashes. The original linker output still aligned individual input sections as needed; this patch fixes the aggregate ELF section-header invariant, but shifting code addresses can also mask a separate layout-sensitive bug. Suggested comment:

```ld
/* Move .text after the boot info and satisfy its current 8-byte
 * output-section alignment requirement. */
_stext = ALIGN(ADDR(.start_block) + SIZEOF(.start_block), 8);
```

With that wording adjustment, I have no issue with the alignment implementation.

chmanie:
Thank you for the fix @kosmar! If you allow edits by maintainers for this PR, we can also take care of rebasing and testing this properly.

kosmar:
@chmanie Thanks — updated the `memory.x` comment and softened the PR description accordingly. Maintainer edits are enabled, so feel free to rebase and test whenever you have a chance.

ArthurGibert (approved):
Reviewed and empirically verified. Built `main` pre-fix and reproduced the exact `rust-lld` alignment warning this PR fixes; built this branch and confirmed the warning is gone, `.text` lands 8-aligned (confirmed via `readelf`), and the resulting release ELF still converts to a valid UF2. No correctness issues.

One minor, non-blocking note: the `ALIGN(..., 8)` value is hardcoded with nothing asserting it matches `.text`'s actual required alignment. If a future toolchain/dependency bump raises that requirement, the same warning would silently reappear — it's a linker warning, not a rustc one, so it isn't caught by `-D warnings` or any CI gate. Not worth blocking on.

ArthurGibert (approved):

ArthurGibert:
Approved on my side. @chmanie let me know

chmanie:
@ArthurGibert yeah fine by me

chmanie (approved):

PR #620, opened 2026-08-03
fix: silence stuck gate in genseq and tb3po

ArthurGibert:
## Summary
- genseq: the clock task never handled `ClockEvent::Stop`, so a legato-held gate (or any gate mid-duty-cycle) stayed high forever once the clock stopped, since no further tick would arrive to resolve it.
- tb3po: the gate/MIDI retrigger block — the only place a slid note's off is ever sent — was skipped entirely while muted, so muting mid-slide left the gate/note stuck until an unmuted gated step eventually came around.

## Test plan
On hardware, for both GenSeq and TB-3PO:
- [ ] Start a pattern with legato/slide active, stop the clock mid-tie, confirm the gate (and MIDI note, if used) goes low/off immediately.
- [ ] Start a pattern with slide active, press mute mid-tie, confirm the gate/note goes low/off immediately instead of staying held.
- [ ] Confirm normal (non-slide) sequencing and mute/unmute behavior is unaffected.

### DIFF
--- a/faderpunk/src/apps/genseq.rs
+++ b/faderpunk/src/apps/genseq.rs
@@ -278,6 +278,20 @@ pub async fn run(
                     }
                     register_pitch = storage.query(|s| s.register_pitch);
                 }
+                ClockEvent::Stop => {
+                    // No phase reset (see ClockEvent::Stop docs) — just silence
+                    // whatever note/gate is currently held, including a legato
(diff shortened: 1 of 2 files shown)


ArthurGibert:
Splitting into two PRs per our merging policy: #621 (genseq) and #622 (tb3po).

PR #623, opened 2026-08-03
feat: Loop de Cay additive overdub loop

kosmar:
## Summary
- New 1ch app **Loop de Cay**: clocked additive overdub loop with per-wrap level decay (kill-floor to true 0)
- Five I/O modes via Config: Pitch→MIDI, Gate→MIDI, MIDI→MIDI, MIDI→CV, Fader→Both
- Gestures: Press = gate/play, Long = mute (no fader move), Shift+Short = arm, Shift+Long = erase
- Alt = decay, Third = 1–8 bar window (crop/reveal + virgin tile); poly 4; RAM-only buffer
- App ID **38** (avoids Hold Sam 36 / Harmonica 37 WIP IDs)

## Test plan
- [ ] Flash firmware and add Loop de Cay on a channel
- [ ] Fader→Both: arm, play short notes with button+fader, confirm loop repeats on clock
- [ ] Overdub adds notes; Alt decay fades velocity to silence; mute pauses decay
- [ ] Shift+Long erases; Third shortens/extends bars (tile on virgin extend, reveal after crop)
- [ ] Spot-check other modes (Pitch→MIDI, Gate→MIDI, MIDI→MIDI, MIDI→CV)

## Prerequisites

- #640 — `ParamStore::update`, so an app can write its own params back to the host

On top of the PRs above there is an app-side change: `ClockEvent::Tick` became
`Tick(u64)` in #579, so the match arm here has to be updated.

#632.

### DIFF
--- a/docs/apps/38-loop-de-cay/manual.json
+++ b/docs/apps/38-loop-de-cay/manual.json
@@ -1,0 +1,89 @@
+{
+  "id": 38,
+  "slug": "loop_de_cay",
+  "name": "Loop de Cay",
+  "status": "wip",
+  "tag": "ready",
+  "description": "Additive overdub loop with level decay",
+  "color": "Violet",
+  "icon": "Sequence",
+  "channels_count": 1,
+  "cv_jack": true,
+  "params": [
+    "Mode",
+    "MIDI Channel",
+    "MIDI In",
+    "MIDI Out",
+    "Base Note",
+    "Span",
+    "Color",
+    "Range",
+    "V/Oct",
+    "Decay Mode"
+  ],
+  "storage": [
+    "Main pitch fader",
+    "Decay strength",
+    "Bars",
+    "Muted",
+    "Armed"
+  ],
+  "columns": [
+    {
+      "sections": [
+        {
+          "heading": "Faders",
+          "items": [
+            "<strong>Main</strong> — Pitch (span from Base Note); spectrum LED meter. In Pitch→MIDI it transposes the CV pitch, so centre = unchanged and an unpatched jack leaves the fader in charge",
+            "<strong>Alt</strong> (Shift+Fader) — Decay strength: bottom = immortal, top ≈ 4 wraps (violet→red)",
+            "<strong>Third</strong> (Btn+Fader; gate modes while muted) — Decay mode zones (IR→UV): Random Rose · Pitch Up Orange · Pitch Down Yellow · Vel+Root Lime · Pitch Root Cyan · Velocity Blue · Gate Violet"
+          ]
+        }
+      ]
+    },
+    {
+      "sections": [
+        {
+          "heading": "Buttons",
+          "items": [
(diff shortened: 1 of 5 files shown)


kosmar:
## Update: Pitch→MIDI fader + Span

`e9822be` — In Pitch→MIDI, Mode 0 always opens a CV in jack and `resolve_pitch` preferred that path, so the fader (and Span) never drove pitch when unpatched. Open CV rests near 0 V and pinned notes to Base Note; Hold+Fader glissando was dead for the same reason.

CV pitch and the bipolar fader offset now stack: unpatched the fader keeps the documented span; with a cable it transposes; centre = 0 semitones (also the saved default).

### Test on hardware
- Loop de Cay · Pitch→MIDI · Base Note C3 · Span 120 · no cable on the jack
- Hold button + move fader → full ±60 semitone range (scale-quantized), not stuck on C3
- Patch V/Oct: centre = unchanged CV pitch; move fader to transpose

kosmar:
Pushed two fixes found while testing on the playground build:

- `fix(loop-de-cay): let a plain hold sustain again` — the Hold+Fader glissando chased its target unconditionally, and since the target includes CV in, an unpatched jack jittering across a quantizer step retriggered the held note. A plain hold now sustains; the glide only engages once the fader has actually moved during the hold.
- `fix(loop-de-cay): end loop voices even when clock ticks are skipped` — note starts and ends tested the loop position for exact equality. The clock gatekeeper publishes ticks immediately, so a subscriber that falls behind drops them rather than stalling the device clock, and a missed end position left the note sounding until its voice slot was stolen. Both now test whether the position was crossed since the last tick, wrap-around included.

Written and pushed by an AI coding agent on behalf of @kosmar. Verified to build; hardware test still pending.

kosmar:

ArthurGibert:

Ran `loop_de_cay.rs` against the actual submission gate (`pr-scope-check.sh`) — clean, no unsafe code, no panics, no MAX11300/storage bypass, no un-yielding loop — and did a real `cargo build` against current `main` with it dropped in, which passed too. It does have 15 `.unwrap()`/`.expect()` calls without a justification comment, which the gate soft-flags for human judgment rather than blocking (same treatment some existing official apps get) — worth a look before resubmitting, not a blocker.

**About `docs/apps/38-loop-de-cay/manual.{json,md}`**: this one's on us, not you — it was listed as a sanctioned option in `CONTRIBUTING.md`/the scope-check script without anyone actually verifying it was backed by real tooling, and it wasn't: nothing in the configurator reads it (confirmed against #606; its entry doesn't show up on the `/#/manual` page at all). You followed the policy as documented — the policy was wrong. Filed as [#656](https://github.com/ATOVproject/faderpunk/issues/656), and [#657](https://github.com/ATOVproject/faderpunk/pull/657) has already removed it as an option here. So `faderpunk-community-apps` isn't just a better fit organizationally — right now it's the only place this content would actually end up visible anywhere.

kosmar:
Moving this to faderpunk-community-apps as requested: https://github.com/ATOVproject/faderpunk-community-apps/pull/12

Closing here. App ID remapped 38 → 109 (community range).

PR #629, opened 2026-08-06
feat: add Venn dual-Euclidean boolean-logic app

kosmar:
## Summary
- New 2ch app **Venn** (ID **40**): dual independent Euclidean layers combined with boolean logic (**OR / AND / XOR / Accnt**)
- Gate-only (no pitch) — distinct from Euclid (single layer) and GenSeq (Turing + pitch CV); inspired by OXI ONE MKII GEN dual-layer Logic
- Latch layers: Main = pulses A/B, Alt = rotation A/B, Third = length A/B (2–32, lengths independent for polyrhythm)
- Shift+Btn0 cycles Logic; Shift+Btn1 toggles post-logic invert (shadow); Btn1 mutes both outs
- Jack0 = primary logic result; Jack1 coupled per mode (overlap / symmetric diff / accent); MIDI Note A/B; Config: Division, GATE %, Prob %, Vel %, Color (Layer B auto-complement)

## Test plan
- [ ] Flash firmware and place Venn on two free channels
- [ ] With clock running: scrub Main pulses A/B; hear/see Jack0/Jack1 and MIDI notes
- [ ] Shift: rotate each layer; Btn0-hold: change lengths independently (XOR/AND get interesting)
- [ ] Shift+Btn0 cycles OR→AND→XOR→Accnt; button LED color follows logic; Shift+Btn1 invert shadow
- [ ] Mute (Btn1); Prob % &lt; 100 adds flakiness; Prob 100 stays deterministic
- [ ] Scene save/load restores pulses/rot/len/logic/einv/mute

## Prerequisites

- #641 — shared `genre_palette` / `groove` / `led_fx` modules

One more dependency is deliberately **not** proposed yet:
`App::wait_while_perf_muted`, which this branch calls. That mechanism does not
exist on `main` at all, and on the integration branch it is used by seven stock
apps — so it needs its own PR and its own review rather than riding along here.
The call will be dropped from this branch until that PR exists.

#632.

### DIFF
--- a/docs/apps/40-venn/manual.json
+++ b/docs/apps/40-venn/manual.json
@@ -1,0 +1,111 @@
+{
+  "id": 40,
+  "slug": "venn",
+  "name": "Venn",
+  "status": "wip",
+  "tag": "ready",
+  "description": "Dual Euclidean layers + boolean logic; scale-aware melody via Extent",
+  "color": "Cyan",
+  "icon": "Euclid",
+  "channels_count": 2,
+  "cv_jack": true,
+  "params": [
+    "MIDI Channel",
+    "MIDI Note A",
+    "Interval B",
+    "MIDI Out",
+    "Division",
+    "GATE %",
+    "Prob %",
+    "Humanize",
+    "Color",
+    "Follow device tonic"
+  ],
+  "storage": [
+    "Pulses A",
+    "Pulses B",
+    "Extent",
+    "Interval",
+    "Length A",
+    "Length B",
+    "Logic",
+    "Invert",
+    "Muted"
+  ],
+  "columns": [
+    {
+      "sections": [
+        {
+          "heading": "Faders",
+          "items": [
+            "<strong>Main</strong> — F0 pulses A · F1 pulses B (density; absolute pulse count, Bjorklund redistributes)",
+            "<strong>Alt</strong> (Shift) — F0 Extent (scale-degree span 0–24 from Note A) · F1 Interval (0–12 diatonic steps above the line for Ch2)",
+            "<strong>Third</strong> (Btn1 or Btn2 hold) — F0 length A · F1 length B (2–32; independent)",
+            "Rotation fixed at 0 (slots freed for Extent / Interval)",
+            "<strong>Division</strong> (config) — note values 4/1 … 1/32T (default 1/16)"
+          ]
+        }
+      ]
+    },
+    {
+      "sections": [
+        {
+          "heading": "Buttons",
+          "items": [
+            "<strong>Btn1 or Btn2 hold</strong> — Third (length A/B on F0/F1)",
+            "<strong>Btn2 tap</strong> (no fader move) — Mute both outs",
+            "<strong>Shift+Btn1</strong> — Cycle Logic OR → AND → XOR → Accnt",
+            "<strong>Shift+Btn2</strong> — Toggle post-logic invert (shadow); white↔none fade like Heat Pump"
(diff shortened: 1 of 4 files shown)


kosmar:
Pushed `fix(venn): close the gate even when clock ticks are skipped`.

Step advance and gate-off both tested the tick number for exact equality. The clock gatekeeper publishes ticks immediately, so a subscriber that falls behind drops them instead of stalling the device clock — and a missed gate-off tick left the MIDI note and the CV gate stuck high until the next one happened to land. The step now fires when the division index changes, and the gate closes on the first tick at or past its offset within the step.

Written and pushed by an AI coding agent on behalf of @kosmar. Verified to build; hardware test still pending.

kosmar:
Follow-up: the logic modes were not audibly distinguishable over MIDI.

Ch2 was derived from the logic result too, so OR, AND and XOR all emitted the same onset set (`a || b`) — the two modes only swapped which voice (line vs. line + Interval) played at a given step. With the default Interval of `m2` that is one scale degree apart, which is effectively inaudible. Ch2 now carries Layer B unchanged, so the mode only affects Ch1. Accnt keeps its accent voice (`a && b`).

Spawn also no longer forces OR. `run()` restarts on every param change, so hardcoding OR there reset the mode and overwrote the stored value whenever anything was edited in the configurator.

Hardware test checklist:

- [ ] Spawn Venn, start the clock, defaults (length 16, pulses ~7 / ~3): both channels fire.
- [ ] Shift + Btn1 cycles OR → AND → XOR → Accnt; each step is clearly audible. AND is noticeably thinner, Accnt follows Layer A only.
- [ ] Raise Interval (Alt + F1) to a fifth: the two voices are clearly separated and the mode differences stand out further.
- [ ] Pick a non-OR mode, then change any param in the configurator: the mode stays put and does not snap back to OR.
- [ ] Save a scene in AND, load another scene and come back: AND is restored.
- [ ] Gate jacks 1 and 2 follow the same pattern as the MIDI notes in every mode.
- [ ] Shift + Btn2 (eInv) still inverts both outputs; Btn2 tap still mutes.

kosmar:

ArthurGibert:

Ran `venn.rs` against the actual submission gate (`pr-scope-check.sh`) — clean, no unsafe code, no panics, no MAX11300/storage bypass, no un-yielding loop — and did a real `cargo build` against current `main` with it dropped in, which passed too. It does have 10 `.unwrap()`/`.expect()` calls without a justification comment, which the gate soft-flags for human judgment rather than blocking (same treatment some existing official apps get) — worth a look before resubmitting, not a blocker.

**About `docs/apps/40-venn/manual.{json,md}`**: this one's on us, not you — it was listed as a sanctioned option in `CONTRIBUTING.md`/the scope-check script without anyone actually verifying it was backed by real tooling, and it wasn't: nothing in the configurator reads it (confirmed against #606; its entry doesn't show up on the `/#/manual` page at all). You followed the policy as documented — the policy was wrong. Filed as [#656](https://github.com/ATOVproject/faderpunk/issues/656), and [#657](https://github.com/ATOVproject/faderpunk/pull/657) has already removed it as an option here. So `faderpunk-community-apps` isn't just a better fit organizationally — right now it's the only place this content would actually end up visible anywhere.

kosmar:
Moving this to faderpunk-community-apps as requested: https://github.com/ATOVproject/faderpunk-community-apps/pull/14

Closing here. App ID remapped 40 → 111 (community range).

PR #599, opened 2026-07-21
fix: isolate MIDI clock ticks from note/CC out queue

kosmar:
## Summary
- Route MIDI realtime clock/transport through a dedicated `MIDI_CLOCK_CHANNEL` so app note/CC bursts cannot fill `MIDI_CHANNEL` and drop ticks (external gear then reads wrong BPM).
- Drain that queue with priority in `midi_out_task`; gatekeeper still uses non-blocking `try_send`.
- Document in `AGENTS.md` that AI-authored commits must say so in the body, and that agents must never merge/approve PRs.

Authored by an AI coding agent on behalf of @kosmar.

## Test plan
- [ ] On hardware: run internal MIDI clock out while apps generate heavy note/CC traffic; confirm external gear BPM stays stable (no dropped ticks / tempo jumps).
- [ ] Start/stop/continue transport still reaches configured MIDI outs.
- [ ] Note/CC from apps still egress normally under light load.

PR #631, opened 2026-08-07
fix: prioritize MIDI realtime output

chmanie:
## Summary
- isolate MIDI timing-clock ticks from app note/CC traffic
- deliver transport through a reliable queue ahead of clock ticks
- bound stale tick backlog while keeping clock generation non-blocking

Supersedes #599.

## Verification

`cargo fmt --all -- --check`, firmware clippy, and `./build-uf2.sh` pass.

## Hardware test
- [ ] Generate heavy note/CC traffic while sending internal clock; confirm stable BPM externally
- [ ] Confirm Start, Stop, Continue, and Reset reach each configured MIDI output
- [ ] Confirm normal note/CC output under light load

### DIFF
--- a/faderpunk/src/tasks/clock.rs
+++ b/faderpunk/src/tasks/clock.rs
@@ -27,7 +27,7 @@ use crate::{
     state::{is_clock_running, update_state},
     tasks::{
         max::{MaxCmd, MAX_CHANNEL},
-        midi::{MidiClockMsg, MidiOutEvent, MIDI_CHANNEL},
+        midi::{MidiRealtimeMsg, MIDI_CLOCK_CHANNEL, MIDI_TRANSPORT_CHANNEL},
     },
     Spawner, GLOBAL_CONFIG_WATCH,
 };
@@ -335,7 +335,8 @@ async fn metronome() {
 #[embassy_executor::task]
 async fn run_clock_gatekeeper() {
     let clock_publisher = CLOCK_PUBSUB.publisher().unwrap();
-    let midi_sender = MIDI_CHANNEL.sender();
+    let midi_clock_sender = MIDI_CLOCK_CHANNEL.sender();
+    let midi_transport_sender = MIDI_TRANSPORT_CHANNEL.sender();
     let clock_in_receiver = CLOCK_IN_CHANNEL.receiver();
     let mut config_receiver = GLOBAL_CONFIG_WATCH.receiver().unwrap();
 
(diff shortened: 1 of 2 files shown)


ArthurGibert (commented):
Splits the shared MIDI output channel into three (notes/CC, clock, transport) so note/CC backlog can no longer starve clock ticks or Stop/Start. This is a genuine, well-targeted fix for real head-of-line blocking.

Hardware validation: flashed on real hardware and stress-tested well beyond this PR's original scope — 16 concurrent LFO apps (several in NRPN mode) continuously flooding the app-level MIDI queue (millions of intentional, by-design drops per `send_cc`'s own drop-on-full-queue policy, confirmed not a bug). Across 1.2M+ log lines under this load, the clock priority channel recorded zero drops (verified via temporary instrumentation), the physical tempo LED stayed visually steady, and external MIDI monitoring confirmed clean clock signal on both USB and DIN outputs throughout. Confirms the PR's core mechanism works: clock delivery isn't starved by note/CC congestion, even at a load far beyond realistic use.

Residual gaps, suggest tracking as follow-up rather than blocking:
1. `midi.rs:127` — `MIDI_CLOCK_CHANNEL` depth 16 lets stale ticks back up during a USB stall specifically, then burst-fire on recovery — contradicts its own doc comment. (Not exercised by this hardware test, which didn't stall the USB host.)
2. `midi.rs:393` — sustained clock/transport traffic can starve `MIDI_CHANNEL`, backpressuring `midi_distributor` and potentially freezing an app task mid-`send_note_on`/`send_cc`.
3. `midi.rs:360` — nothing flushes `MIDI_CLOCK_CHANNEL` on Stop/Start/Reset, so stale ticks can play out after Stop.
4. `clock.rs:463` — the transport channel's blocking send reproduces the same stall-propagation bug this PR fixes, just relocated — a full transport channel now stalls CV clock output too.
5. Minor: `midi.rs:357`'s manual `try_receive` pre-check is redundant with `select4`'s own fixed poll order.
6. Minor, pre-existing: `write_msg_to_uart0` has no timeout, unlike the USB path.

A related investigation (an unconfirmed user jitter report) additionally found that once the dispatcher commits to a write it isn't preemptible — up to ~28ms worst case for an NRPN write under a USB stall — and produced a fix for that plus #2 above, now on branch `fix/midi-jitter-on-realtime-priority` (pushed, not yet a PR).

Recommend merging as-is given the hardware validation above, and tracking the remaining gaps as documented follow-up.

ArthurGibert (approved):

PR #632, opened 2026-08-07
fix: support current Rust toolchains

chmanie:
## Summary
- migrate package-local build-std configuration to the current nightly immediate-abort panic strategy
- keep the workspace manifest compatible with stable Cargo used by CI and releases
- fix the current-nightly MIDI packet Clippy lint

Supersedes #611.

## Verification
- latest nightly: formatting, firmware/libfp Clippy, libfp tests, firmware release build, and UF2 generation
- latest stable: formatting, firmware/libfp Clippy, libfp tests, and firmware release build

## Hardware test
- [ ] Flash firmware and confirm the firmware boots normally

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -70,32 +70,32 @@ Follow this for **every** task, in order. Do not skip steps.
 
 ## Environment & Toolchain
 
-The toolchain (Rust nightly, Node, pnpm, picotool, probe-rs) is provided by
+Project commands use Rust, Node, pnpm, picotool, and probe-rs through
 **devenv** via **direnv** (`devenv.nix`, `.envrc`). A plain shell does not have
-it on PATH. Run project commands through the environment:
+the complete toolchain on PATH. Run project commands through the environment:
 
 ```bash
 direnv exec . <command>      # preferred (needs `direnv allow` once)
 devenv shell -- <command>    # equivalent fallback
 ```
 
-The firmware targets `thumbv8m.main-none-eabihf` and uses nightly features
-(`build-std`). `faderpunk/.cargo/config.toml` sets this target by default, so a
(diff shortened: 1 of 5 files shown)


ArthurGibert (commented):
Migrates `panic_immediate_abort` to the syntax current nightly requires, and swaps `chunks_exact(4)` → `as_chunks::<4>()` in `midi.rs`. Confirmed necessary (the old config fails to compile `core` on current nightly — reproduced directly) and confirmed correct (`as_chunks` behavior is identical to `chunks_exact`, zero-cost per asm diff; the new `alloc` build-std addition is genuinely required by `postcard-bindgen-core`, not scope creep).

Worth addressing before merge: the new config hard-fails on stable Cargo, where the old one silently no-op'd and still produced a working (larger) binary — reproduced directly both ways. There's no `rust-toolchain.toml` anywhere in the repo pinning nightly for the `faderpunk` package, and `gen-bindings.sh` has to say `cargo +nightly` explicitly, implying bare `cargo` resolves to stable here. That risks breaking the documented `cd faderpunk && cargo build --release` / `./build-uf2.sh` workflow for anyone without a personal rustup override already set. Suggest adding a `rust-toolchain.toml` pinning nightly for `faderpunk/`.

Worth knowing, not blocking:
- Because `.cargo/config.toml` is scoped to `faderpunk/`, CI/release builds (which run from repo root) never see this fix at all — this PR only affects the local `build-uf2.sh` dev path.
- `panic = "immediate-abort"` makes ordinary app panics compile to silent traps with no defmt/RTT output — pre-existing, not introduced here, but worth a code comment given how much it undermines local debugging.
- A release-build `rust-lld` alignment warning showed up during testing — almost certainly the same pre-existing issue #614 fixes independently, not new here.

ArthurGibert (approved):

PR #636, opened 2026-08-10
fix: never block the clock gatekeeper on a full pubsub

kosmar:
The clock gatekeeper publishes ticks with `publish().await`. A subscriber that
sleeps while holding its slot (the metronome does, for the beat flash) lets the
shared `CLOCK_PUBSUB` queue fill up. Once it is full the gatekeeper blocks,
`CLOCK_IN` backs up behind it, and the whole device clock stops — apps go silent
after a few bars while USB and the UI still look healthy.

Ticks now use `publish_immediate`, so a lagging subscriber loses a beat instead
of stalling the publisher. The analog tick/reset writes switch from
`send().await` to `try_send` for the same reason: a full MAX queue must not be
able to pin the gatekeeper either.

Transport events (Start/Stop/Reset) keep `publish().await`. They are rare,
one-shot, and worth delivering reliably; with ticks no longer able to wedge the
publisher, waiting there is bounded.

Note: #631 touches the same two files (it adds the dedicated realtime queue) but
does not change the gatekeeper's blocking publish, so the two are complementary.
Whichever lands second needs a small rebase.

## Test plan

- [ ] Run a dense layout (several clocked apps at fine divisions, e.g. 1/16-1/32) for
      a few minutes on hardware and confirm the clock keeps running and apps keep
      sounding.
- [ ] Confirm the Scene LED beat flash still tracks the beat.
- [ ] With an AUX jack set to Clock Out / Reset Out, confirm analog ticks and reset
      pulses still fire.
- [ ] Start/Stop/Continue from an external clock source still behave.

### DIFF
--- a/faderpunk/src/tasks/clock.rs
+++ b/faderpunk/src/tasks/clock.rs
@@ -275,10 +275,11 @@ async fn send_analog_ticks(spawner: &Spawner, config: &GlobalConfig, counters: &
         }
     }
     if !ports.is_empty() {
-        MAX_CHANNEL
+        // try_send for the same reason: a full MAX queue must not stall the
+        // clock gatekeeper. A dropped analog tick is better than a dead clock.
+        let _ = MAX_CHANNEL
             .sender()
-            .send(MaxCmd::GpoSetHighMany(ports.clone()))
-            .await;
+            .try_send(MaxCmd::GpoSetHighMany(ports.clone()));
         spawner.spawn(analog_tick_release(ports, 5)).ok();
     }
 }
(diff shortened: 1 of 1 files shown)


ArthurGibert:
Reviewed and pushed a follow-up fix: the `GpoSetLowMany` send in `analog_tick_release` was switched to `try_send`, but that task runs off the gatekeeper's execution path (it's spawned separately), so it never needed to be non-blocking — doing so meant a momentarily full `MAX_CHANNEL` would silently drop the "set low" pulse and leave a ClockOut/ResetOut jack stuck high indefinitely, with no retry until the next Start/Reset.

Reverted just that one send back to blocking `.send().await` (`c97cb7aa`); `send_analog_ticks`/`send_analog_reset`'s `try_send` calls stay as-is since those genuinely are on the gatekeeper's blocking path.

Verified on hardware with an instrumented A/B: flooded `MAX_CHANNEL` continuously while running the clock at 900 BPM.
- Pre-fix (`try_send` in `analog_tick_release`): 13,280 dropped releases (port left HIGH) over ~42s.
- With this fix: 0 dropped releases over ~32s of the same flood, and the blocking send never waited more than 1ms for a queue slot even under sustained saturation.
- Clock tick progression stayed continuous throughout both runs (no stalls), confirming the `publish_immediate` part of this PR does what it's meant to do.

Also flashed and manually tested the fixed build without the synthetic flood — works well.

kosmar:
Thanks for catching the `analog_tick_release` `try_send` and for the hardware A/B — that stuck-HIGH case would have been nasty in the field. Looks good to merge from our side.

PR #637, opened 2026-08-10
feat: report the panic site over MIDI

kosmar:
Draft: needs a decision against #632 first — see "Interaction with #632" below.

When Core 1 aborts, the device still looks healthy. Core 0 keeps USB enumerated
and the MIDI clock ticking, so the only symptom is that app notes stop. Without
a debug probe attached there is no way to find out where it died.

This replaces `panic-probe` with a small handler that logs the panic site over
defmt (as before, minus the message text) and records `file`/`line` for a Core-0
task, which re-sends them as CCs on MIDI channel 16 once per second. A plain MIDI
capture then carries the panic site. `scripts/decode-panic-beacon.py` maps the
file hash back to a path.

This is how the Core-1 abort behind the recent "app goes silent after a few
notes" reports was located: the beacon pointed straight at the offending line.

Size: 951,900 bytes of flash vs 949,408 on main, so +2,492 bytes. Dropping
`panic-probe` saves about 20 KiB; keeping a defmt line for probe users spends
most of it back.

## Interaction with #632

#632 sets `panic = "immediate-abort"`, which compiles the `#[panic_handler]`
away and makes this beacon inert. The two are not textually in conflict (this
branch does not touch `Cargo.toml` or `.cargo/config.toml`), but they pull in
opposite directions and want sequencing:

- If #632 lands first, this needs `panic = "abort"` to do anything, which gives
  back the flash #632 saves.
- If this lands first, #632 silently disables it.

Worth deciding whether post-mortem visibility without a probe is worth a
non-immediate abort. Happy to close this if the answer is no.

Note: `cargo clippy` on current nightly fails on `midi.rs:520` (`chunks_exact`)
on main as well; #632 fixes that.

## Test plan

- [ ] Flash and confirm normal operation is unchanged (apps, MIDI, clock).
- [ ] Provoke a panic in an app on Core 1, then capture MIDI and confirm CC 110-115
      arrive on channel 16 once per second.
- [ ] Run `scripts/decode-panic-beacon.py` against the capture and confirm it prints
      the correct file and line.
- [ ] With a probe attached, confirm the panic site still shows up over RTT.

### DIFF
--- a/faderpunk/src/main.rs
+++ b/faderpunk/src/main.rs
@@ -19,6 +19,7 @@ mod version {
     include!(concat!(env!("OUT_DIR"), "/version.rs"));
 }
 
+use defmt_rtt as _;
 use embassy_executor::{Executor, Spawner};
 use embassy_rp::clocks::{ClockConfig, CoreVoltage};
 use embassy_rp::config::Config;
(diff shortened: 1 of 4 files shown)


kosmar:
#632 landed the other way: it removed `immediate-abort` rather than setting it, so the beacon would work on current `main`.

Still interested, or should I close this?

PR #639, opened 2026-08-10
feat: add non-blocking note and CC senders for voice engines

kosmar:
`send_note_on` / `send_note_off` await space in the app MIDI queue. For a voice
engine that decides notes inside a clock step that is the wrong trade: the task
parks on Core 1, pressure backs up into USB TX, and every later note is delayed
too. Losing a single note is cheaper.

Adds sync, non-blocking variants alongside the existing ones:

- `try_send_note_on` / `try_send_note_off` — drop instead of awaiting
- `try_send_cc` — same payload as `send_cc` (already non-blocking), but callable
  from sync code

Being non-async matters as much as being non-blocking: these fire from closures
inside tight clock loops where `.await` is not available.

Nothing existing changes behaviour — the blocking senders stay as they are, and
apps opt in. The `#[allow(dead_code)]` is because the first callers are the WIP
app branches (#634, #633, #604, #609), which cannot compile against `main`
without this.

Note for review: `cargo clippy -D warnings` currently fails on `main` itself
with a `chunks_exact` lint in `tasks/midi.rs` (toolchain drift, unrelated to
this change).

**Update after review:** `try_send_note_off` has been dropped — a dropped NoteOff has no future message to correct it, and `APP_MIDI_CHANNEL` is shared across all 16 channels, so the drop is not something a voice engine can avoid. NoteOff goes through the blocking `send_note_off` from the async voice tasks. Only `try_send_note_on` and `try_send_cc` remain.

### DIFF
--- a/faderpunk/src/app.rs
+++ b/faderpunk/src/app.rs
@@ -448,6 +448,12 @@ impl MidiOutput {
     /// notes inside a tight clock step want the opposite trade: lose one note
     /// rather than delay every later one. Being sync also lets them fire from
     /// closures where `.await` is not available.
+    ///
+    /// There is deliberately no `try_send_note_off` counterpart: a dropped
+    /// NoteOn is silence, but a dropped NoteOff is a note that hangs until
+    /// something else happens to end it, and `APP_MIDI_CHANNEL` is shared by
+    /// all 16 channels, so a busy neighbour can cause the drop. Send the
+    /// matching NoteOff with [`Self::send_note_off`] from an async voice task.
     #[allow(dead_code)]
     pub fn try_send_note_on(&self, note_number: MidiNote, velocity: u16) {
         let event = LiveEvent::Midi {
@@ -455,23 +461,6 @@ impl MidiOutput {
             message: MidiMessage::NoteOn {
                 key: note_number.into(),
                 vel: scale_bits_12_7(velocity),
-            },
-        };
-        let msg = MidiMsg::new(event, self.midi_out, MidiEventSource::Local);
-        let _ = self.midi_sender.try_send((self.start_channel, msg));
-    }
-
-    /// Non-blocking, non-async NoteOff — see [`Self::try_send_note_on`].
-    ///
-    /// Pair it with `try_send_note_on`: an engine that can drop a NoteOn must
-    /// be able to drop the matching NoteOff without stalling.
-    #[allow(dead_code)]
-    pub fn try_send_note_off(&self, note_number: MidiNote) {
-        let event = LiveEvent::Midi {
-            channel: self.midi_channel,
-            message: MidiMessage::NoteOff {
-                key: note_number.into(),
-                vel: 0.into(),
             },
         };
         let msg = MidiMsg::new(event, self.midi_out, MidiEventSource::Local);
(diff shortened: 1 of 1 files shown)


ArthurGibert (commented):
`try_send_note_on` and `try_send_cc` are fine, but `try_send_note_off` risks stuck notes: if a NoteOff drops after its NoteOn already reached the synth, there's no future update to correct it — unlike CC, which `send_cc`'s own doc comment already flags ("unlike note on/off, dropping here can't leave a stuck note").

`APP_MIDI_CHANNEL` is one global queue shared across all 16 channels, so this isn't confined to a voice engine overloading itself — any busy neighbor app can cause the drop.

Nothing adopts this yet, but #634/#633/#604/#609 are the named targets, and the doc comment currently steers them toward pairing it with `try_send_note_on`. Suggest either dropping `try_send_note_off` from this PR (fall back to blocking `send_note_off`), or giving NoteOff a reserved queue lane — same tradeoff #631 already solved for clock/transport.

kosmar:

Agreed, and dropped — `47a2467b` removes `try_send_note_off`. `try_send_note_on` and `try_send_cc` stay, and the `try_send_note_on` doc now says explicitly that there is deliberately no NoteOff counterpart and that the matching NoteOff goes through the blocking sender.

What settled it: the “non-async matters as much as non-blocking” argument in the PR description turns out not to apply to NoteOff. Every real caller on the WIP branches sits in a dedicated async voice task (`let voice = async { … }` in Arp, `fut_voice` in Bassment and Contura), not in a sync closure inside the clock loop — that separation was the whole point of isolating the voice engines. So `.await` is available at every one of those call sites, and falling back to `send_note_off` costs at most a late NoteOff in that one app's voice loop. It cannot stall the clock subscriber. Contura even ended up hand-rolling the safe version: it checks the return value, keeps the note in `note_on` and retries on the next 1 ms pass — which is just a worse `send_note_off().await`.

Your point about `APP_MIDI_CHANNEL` being one global queue is the part that makes this non-negotiable, though. If the drop were self-inflicted, an app could at least be written to avoid it. It isn't, so no amount of care in the voice engine helps.

On the reserved NoteOff lane: I think that's the better long-term answer, not just an equivalent alternative. Blocking on `send_note_off` fixes correctness by making the voice loop wait, which is fine at today's density but is still the wrong shape — the message that must never be lost is queued behind bulk CC traffic from fifteen other channels. A reserved lane, the same way #631 solved it for clock and transport, gives NoteOff a guarantee instead of a fallback and lets voice engines go fully non-blocking again. I'd rather do that as its own change with its own hardware verification than bolt it onto this PR, so this one stays “add two non-blocking senders”. Happy to open the follow-up issue if you want it tracked.

The WIP branches (#634, #633, #604, #609 and the others that adopted it) still call `try_send_note_off` and will be moved over to `send_note_off().await` before they come up for review.

kosmar:
testing

ArthurGibert:
Reviewed and rebased onto current main. Nice PR — the reasoning is sound, especially dropping `try_send_note_off` after your own review (matches the same "dropped NoteOn = silence, dropped NoteOff = stuck note" tradeoff we independently worked through while auditing a similar MIDI-blocking scenario in another app).

One small cleanup: `try_send_cc` duplicated `send_cc`'s NRPN/Controller message-construction logic instead of `send_cc` delegating to it (it's already fully non-blocking under the hood, just declared `async`). Same duplication pattern between `try_send_note_on` and the existing NoteOn path. Fixed by having `send_cc` call `try_send_cc` directly, and factoring the shared `LiveEvent`/`MidiMsg::new` wrapping into a small private `wrap_midi_msg` helper used by both `send_midi_msg` and `try_send_note_on` — no behavior change, just removes the risk of the two copies drifting apart on a future encoding change.

Builds and lints clean. Looks good to merge.

PR #641, opened 2026-08-12
feat: add shared app helper modules and libfp clock/brightness helpers

kosmar:
## Summary

Adds small shared helpers that the upcoming WIP apps all need, so each app PR
can shrink to its own app file plus a registry line.

Under `faderpunk/src/apps/`, for the genre-based apps (Grooves, Chord Vamp,
Bassment):

- `genre_palette.rs` — shared genre names and 8-bar tropes
- `groove.rs` — swing/feel curves and swing delay maths
- `led_fx.rs` — spectrum colour helpers for the genre scrub axis
- `follow_key.rs` — shared helper for following the device tonality

In `libfp::utils`, for the clocked and LFO-style apps (Super LFO, Manifold,
Venn, Contura, Hold Sam):

- `CLOCK_DIVISIONS` and `division_at` — one division table, and an even mapping
  from a fader to a prefix of it, replacing the per-app tables and the ad-hoc
  `value / 500` style lookups
- `lfo_step` / `lfo_step_modulated` — the exponential rate curve, with and
  without CV modulation
- `quant_step` — phase increment for a clock-synced cycle
- `signal_brightness` — the LED brightness curve for a unipolar or bipolar
  signal

Each of those apps currently carries its own copy on its feature branch, and
the copies have already diverged — `genre_palette` gained `GENRE_PROG_8` on
some branches while others still ship the now-unused `GENRE_COLORS` table.
Landing the helpers once, ahead of the apps, gives them a single source of
truth. The contents are taken from the integration branch these apps are
flashed from, which is ahead of every feature branch for `groove.rs`.

Centralising `signal_brightness` also fixed a bug the copies shared: at full
scale a bipolar value produced `(4095 - 2047) / 8 = 256`, which wrapped to `0`
in `u8` and dimmed the LED instead of lighting it. The shared version clamps
before the cast, and the unit tests cover it.

## Scope

Additive. New app-side modules plus their `mod` declarations, and new functions
plus unit tests in `libfp::utils`. `register_apps!` is untouched and no existing
app or platform file is modified, so stock app behaviour cannot be affected.

The app-side modules are dead code until the first consumer app lands, hence
the `#[allow(dead_code)]` on the declarations; the first app PR to use them can
drop it.

## Test plan

No hardware checklist: this change adds no reachable code path and no runtime
behaviour. `cargo build --release` and `cargo test --lib -p libfp` against
current `main` are the whole verification.

### DIFF
--- a/faderpunk/src/apps/follow_key.rs
+++ b/faderpunk/src/apps/follow_key.rs
@@ -1,0 +1,104 @@
+//! Shared "follow the device tonality" helpers for note-generating apps.
+//!
+//! The device has one global Key + Tonic, reachable live by holding the Scene
+//! button (Fader 4 = Scale, Fader 5 = Tonic). An app that follows the Tonic
+//! turns that fader into a **transpose** — and every following app moves
+//! together, which is what you want when a bass and a melody play the same
+//! tune. Apps expose this as two `Param::bool`s named "Follow device tonic"
+//! and "Follow device scale".
+//!
+//! Scale and Tonic are deliberately separate: an app's Scale is often part of
+//! its own character (Contura's Folk set, Bassment's genre feel), so the usual
+//! default is to follow the Tonic but keep the local Scale.
+//!
+//! **Cost:** every call here copies the whole `GlobalConfig`. Resolve once per
+//! bar or per phrase and cache it — never per step and never per note.
+
+// Each app needs a different part of this surface, and the app feature branches
+// carry one app each — so on those, some of it is unused by construction.
+#![allow(dead_code)]
+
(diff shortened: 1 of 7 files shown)


kosmar:

`48febdc4` adds `device_swing_permille` and `device_swing_reverses` on `groove.rs`.

Apps that layer their own swing on the device clock had no way to ask how much the clock had already moved a step, so they could only switch their own timing off whenever global swing was non-zero. The new helpers report that displacement as a fraction of the app’s grid step (zero when the grid is coarser than the swing window half) and which parity to delay so Feel leans with the clock. Consumed by #633 and #648.

PR #642, opened 2026-08-12
feat: add Manifold, a CV distributor with comparator outputs

kosmar:
Manifold (WIP ID 43) takes one CV input and drives three independently shaped
outputs. Each output has its own attenuverter, offset and slew spread across the
three fader layers, and can instead run as a comparator that emits a gate or a
trigger.

It fills two gaps in the current app set: no app could send one CV to more than
one destination, because every jack is either an input or an output, and no app
could derive a trigger from a continuous voltage — even though most generative
apps expose a Reset trigger input.

Channel 0 is the CV input, channels 1–3 are outputs B, C and D. The comparator
threshold defaults to 2458, the same constant the other WIP apps use for their
own trigger inputs, so a Manifold trigger and a Reset input agree out of the box.

## Test checklist (hardware)

- [ ] CV mode: input follows to all three outputs; per-output attenuverter
      inverts below fader centre and reaches zero exactly at centre
- [ ] Shift+fader offsets each output independently; button+fader slews it
- [ ] Fader 0 trims the input and Shift+Fader 0 offsets it for all three branches
- [ ] Button 0 tap freezes the input; all three outputs hold
- [ ] Button 1–3 tap mutes that output only; scrubbing the fader while the
      button is held does not toggle mute
- [ ] Buttons work independently: hold button 1 and scrub, then tap button 2
- [ ] Gate mode: output goes high above the threshold and only releases below
      threshold minus hysteresis; minimum gate length is honoured
- [ ] Trigger mode: one pulse per rising edge, no burst on a slowly rising ramp
- [ ] Patch a Manifold trigger into another WIP app's Reset input
- [ ] In Range 0–10 V accepts a unipolar source (AD Envelope) without clipping
- [ ] Scene save/load restores all values; outputs sit on saved values right
      after a layout change instead of gliding to them


kosmar:
Branch updated: the internal morph LFO fallback and the mode-keyed LED hues are now on `feat/manifold`.

**Depends on three open PRs**, merged into this branch so it builds on its own:

- #641 — `led_fx::hsv_to_rgb` plus `division_at` / `lfo_step` / `signal_brightness` in `libfp::utils`
- #644 — `App::clock_ticker`
- #640 — `ParamStore::update`

Their commits show up in this diff until they land on `main`; once they do, the diff reduces to Manifold's own files. `morph.rs` is new here (the wave sampler extracted from Super LFO); its colour helpers carry `#[allow(dead_code)]` because Super LFO lands separately. The WIP cheatsheet is deliberately not included — that belongs to #613.

Gates on this branch: `cargo fmt --check`, firmware clippy, `libfp` clippy, and `cargo test --lib -p libfp` (113 passed) all clean.

kosmar:

Manifold now mirrors all four channels as MIDI CC (`7a5c1f5d`).

It was jack-and-fader only, so nothing it did could be seen on a USB scope — the one app in the chain whose shaping you most want to watch was invisible. It sends the conditioned input and the three outs on base CC plus 0..3, taking the value the jack actually carries, so Gate and Trigger read as their digital state and mutes show through.

The loop runs at 1 kHz against the LFO apps’ 125 Hz, and four channels multiply that further, so sends are paced and rotated one channel per 8 ms. That holds the bus at roughly 125 messages/s in total while each channel still refreshes far faster than a scope can draw. Presets saved before the mirror keep ten values and stay silent on MIDI.

Note on the base: this pulls in #639 for `try_send_cc`, the same way the branch already merges #641, #640 and #638. Worth merging #639 first.

kosmar:
## Update

Mode cycle on B/C/D is deferred to button-up so a Third-layer fader scrub (btn+fader) cancels the mode change — same cancel path as mute / Ripppple Process.

kosmar:
Pushed: live mode hues via `glob_modes` (fixes LED hue stuck after mode cycle) + 30° palette `200→320`.

kosmar:

ArthurGibert:

Ran `manifold.rs` against the actual submission gate (`pr-scope-check.sh`) — clean, no unsafe code, no panics, no MAX11300/storage bypass, no un-yielding loop — and did a real `cargo build` against current `main` with it dropped in, which passed too. It does have 16 `.unwrap()`/`.expect()` calls without a justification comment, which the gate soft-flags for human judgment rather than blocking (same treatment some existing official apps get) — worth a look before resubmitting, not a blocker.

**About `docs/apps/43-manifold/manual.{json,md}`**: this one's on us, not you — it was listed as a sanctioned option in `CONTRIBUTING.md`/the scope-check script without anyone actually verifying it was backed by real tooling, and it wasn't: nothing in the configurator reads it (confirmed against #606; its entry doesn't show up on the `/#/manual` page at all). You followed the policy as documented — the policy was wrong. Filed as [#656](https://github.com/ATOVproject/faderpunk/issues/656), and [#657](https://github.com/ATOVproject/faderpunk/pull/657) has already removed it as an option here. So `faderpunk-community-apps` isn't just a better fit organizationally — right now it's the only place this content would actually end up visible anywhere.

kosmar:
Moving this to faderpunk-community-apps as requested: https://github.com/ATOVproject/faderpunk-community-apps/pull/17

Closing here. App ID remapped 43 → 114 (community range).

PR #644, opened 2026-08-13
feat: add clock_ticker for apps that only need the tick count

kosmar:
Split out of #638 at Arthur's request: `clock_ticker` is additive API and has nothing to do with the pubsub/metronome fix.

`App::clock_ticker()` hands an app a plain `fn() -> u64` that reads the current 24 PPQN tick out of `TICK_COUNTER`. Apps that derive everything from the tick number and never react to Start/Stop/Reset can use it instead of `use_clock()`: it costs no `CLOCK_PUBSUB` subscriber slot, and an app that falls behind cannot back up the shared clock queue. `u64::MAX` means “not started”; the counter going backwards means the clock was restarted or reset.

It is `#[allow(dead_code)]` on `main` — the first users are the WIP app branches, where a clocked app that also awaits a MAX write or a MIDI send inside its loop is exactly the subscriber that used to stall the gatekeeper.

Stacked on #638 (which is itself stacked on #636), because `TICK_COUNTER` is introduced there. Cross-repo PRs can only target `main`, so this diff shows all three commits; merge #636 and #638 first and it shrinks to the `app.rs` change alone.

No test checklist: this adds an unused accessor and changes no behaviour on its own.

_Opened by an AI coding agent on @kosmar's behalf._

### DIFF
--- a/faderpunk/src/app.rs
+++ b/faderpunk/src/app.rs
@@ -23,7 +23,7 @@ use crate::{
     events::{EventPubSubChannel, InputEvent},
     tasks::{
         buttons::{is_channel_button_pressed, is_shift_button_pressed},
-        clock::{ClockSubscriber, CLOCK_PUBSUB},
+        clock::{ClockSubscriber, CLOCK_PUBSUB, TICK_COUNTER},
         global_config::get_global_config,
         i2c::{I2cLeaderMessage, I2cLeaderSender},
         leds::{set_led_mode, LedMode, LedMsg},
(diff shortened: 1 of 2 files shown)


ArthurGibert:
I rebased on main so that the merge of this PR doesn't depend on the approval of the other PRs.

PR is justified and applications are real. That's an approval as far as I'm concerned.

@chmanie Let me know what you think

ArthurGibert (approved):

ArthurGibert:
This was merged by mistake — merging a stacked PR directly onto `main` pulled in the full 3-commit stack (#636, #638, and this one) in a single squash commit, without #636/#638 going through their own review and merge. It's been reverted from `main` (force-pushed back to the pre-merge commit). Reopened with identical content as #646.

PR #646, opened 2026-08-14
feat: add clock_ticker for apps that only need the tick count

ArthurGibert:
Reopens #644, which was merged by mistake on 2026-08-14 and then reverted from `main` (force-push reset to the pre-merge commit). Same content, unchanged: three stacked commits (`fix/clock-gatekeeper`, `feat/clock-ticker`, `feat/app-clock-ticker`), since `TICK_COUNTER` is introduced further down the stack.

See #644 for the original description and review discussion.

### DIFF
--- a/faderpunk/src/app.rs
+++ b/faderpunk/src/app.rs
@@ -23,7 +23,7 @@ use crate::{
     events::{EventPubSubChannel, InputEvent},
     tasks::{
         buttons::{is_channel_button_pressed, is_shift_button_pressed},
-        clock::{ClockSubscriber, CLOCK_PUBSUB},
+        clock::{ClockSubscriber, CLOCK_PUBSUB, TICK_COUNTER},
         global_config::get_global_config,
         i2c::{I2cLeaderMessage, I2cLeaderSender},
         leds::{set_led_mode, LedMode, LedMsg},
@@ -795,6 +795,18 @@ impl<const N: usize> App<N> {
         Clock::new()
     }
 
+    /// Poll the current tick instead of subscribing to the clock.
+    ///
+    /// For apps that derive everything from the tick number and never react to
+    /// Start/Stop/Reset. Unlike [`Self::use_clock`] this costs no subscriber
+    /// slot, and an app that falls behind cannot back up the shared clock
+    /// queue. Returns `u64::MAX` before the first tick; the counter going
+    /// backwards means the clock was restarted or reset.
+    #[allow(dead_code)] // first users are the WIP app branches
+    pub fn clock_ticker(&self) -> fn() -> u64 {
+        ticks
+    }
+
     pub fn use_quantizer(&self, range: Range, vpo: VoltPerOct, bypass: bool) -> Quantizer {
         Quantizer::new(range, vpo, bypass)
     }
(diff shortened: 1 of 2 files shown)


chmanie (commented):
The ticks are now part of the clock events that's why it was removed earlier. See here: https://github.com/ATOVproject/faderpunk/blob/25546b547cab9340d0710c52d564e730adcb62d1/faderpunk/src/app.rs#L317

ArthurGibert:
I guess as a work-around one could still subscribe to clock events and use a global variable to share the count to the other loops. Now we just need to decide if that's how we want this case to be handled @chmanie .


For context PR #644 was meant to bring back the ticker for apps that do not use the clock events:

```Split out of https://github.com/ATOVproject/faderpunk/pull/638 at Arthur's request: clock_ticker is additive API and has nothing to do with the pubsub/metronome fix.

App::clock_ticker() hands an app a plain fn() -> u64 that reads the current 24 PPQN tick out of TICK_COUNTER. Apps that derive everything from the tick number and never react to Start/Stop/Reset can use it instead of use_clock(): it costs no CLOCK_PUBSUB subscriber slot, and an app that falls behind cannot back up the shared clock queue. u64::MAX means “not started”; the counter going backwards means the clock was restarted or reset.

It is #[allow(dead_code)] on main — the first users are the WIP app branches, where a clocked app that also awaits a MAX write or a MIDI send inside its loop is exactly the subscriber that used to stall the gatekeeper.

Stacked on https://github.com/ATOVproject/faderpunk/pull/638 (which is itself stacked on https://github.com/ATOVproject/faderpunk/pull/636), because TICK_COUNTER is introduced there. Cross-repo PRs can only target main, so this diff shows all three commits; merge https://github.com/ATOVproject/faderpunk/pull/636 and https://github.com/ATOVproject/faderpunk/pull/638 first and it shrinks to the app.rs change alone.

No test checklist: this adds an unused accessor and changes no behaviour on its own.

Opened by an AI coding agent on @kosmar's behalf.```

**Comment by ArthurGibert** [comment]
Following up after reviewing #636/#638: this PR's `App::clock_ticker()` depends entirely on `TICK_COUNTER`, which no longer exists — #638 replaced it with `METRONOME_SIGNAL` (an `embassy_sync::Signal`), since that's a better fit for its actual single-consumer use case (the metronome) and avoids an always-on 1ms poll. So this PR can't be rebased onto the current stack as-is; it needs a rewrite, not just a merge.

Beyond that, holding off on this makes sense independent of the rebase issue: there's currently no app in `main` calling `clock_ticker()` (it ships `#[allow(dead_code)]`), so merging it now would be unused public API ahead of a real consumer. Recommendation: for now, WIP apps that only need the tick count should use the app-level workaround (a local atomic fed from `ClockEvent::Tick`, same shape this PR was formalizing) rather than wait on this.

If/when a real app needs this for real, worth reconsidering `embassy_sync::watch::Watch` instead of a bare atomic at that point — it's already the codebase's established "shared state, multiple independent readers" primitive (same family as `LAYOUT_WATCH`/`GLOBAL_CONFIG_WATCH`), and unlike a single-consumer `Signal` it cleanly supports several apps polling independently, which is probably the actual shape needed here. Leaving this open for now rather than closing, in case that's useful reference when we get there.

**Comment by ArthurGibert** [comment]
you can use a loop like this until we revisit this:

```let clock_handler = async {
    loop {
        match clk.wait_for_event(ClockDivision::_1).await {
            ClockEvent::Tick(tick) => {
                glob_ticks.set(tick);
            }
            _ => {}
        }
    }
};
```

kosmar:
Thanks — we'll use the drain-loop workaround on the WIP apps (dedicated `wait_for_event` task that only stores the absolute tick; the rest of the app may park on MAX/MIDI). No rush on Watch / `clock_ticker` until something on `main` actually needs it.

PR #650, opened 2026-08-15
feat: add Ripppple, a cumulative waveshaper chain

kosmar:
Ripppple (WIP ID 46) is a cumulative waveshaper chain, not a cascade of stage LFOs. Channel 0 is the root; channels 1–3 each reshape the cumulative result of everything before them. Output D has therefore passed through every preceding stage.

- **Ch0** — root. Jack is a CV input with attenuverter conditioning (Main = trim, Alt = offset, Third = slew). After ~1.2 s of a static jack the input falls back to an internal morphing LFO; those same three fader layers then become speed, morph and skew. A short tap freezes the root value (the value itself, so stepped morph nodes hold too). Shift + short toggles clock sync while the internal LFO runs.
- **Ch1–Ch3 (B/C/D)** — no oscillators of their own. Each stage waveshapes its predecessor: Main = amount, Alt = offset, Third = slew. At amount zero, Fold passes through at unity gain; Soft and Rect blend from dry towards their curve. Long press cycles process Fold → Soft → Rect (same as configurator params Process B/C/D). Shift + long cycles out range ±5 V ↔ 0–10 V (1× / 2× button flash). Short tap mutes with a click-free ramp; fader scrub while held cancels mute or process cycle on release.
- **LEDs** — bipolar meter of each stage’s cumulative output. Five equidistant hues: Ch0 CV in vs internal LFO, then Fold / Soft / Rect on the stages (Ch0 CV/LFO hues shared with Manifold).

Depends on #642 — Ripppple uses the shared `morph` sampler that lands with Manifold, so this branch is cut from `feat/manifold` and should be reviewed and merged after it.

## Test checklist (hardware)

- [ ] Unpatched / static Ch0 falls back to the internal LFO within ~1.2 s; jack movement switches back to CV in immediately
- [ ] Ch0 CV layers: Main = attenuverter trim, Alt = offset, Third = slew; after LFO fallback the same layers are speed / morph / skew
- [ ] Short tap on Ch0 freezes the root value (chain stops); Shift + short on Ch0 toggles clock sync while the LFO runs; stopped transport parks the wave
- [ ] B shapes the root, C shapes B, D shapes C — changing an upstream amount/process visibly changes only that stage and everything after it
- [ ] Amount at zero: Fold is unity thru; Soft and Rect are dry; raising amount applies the curve
- [ ] Long press on B/C/D cycles Fold → Soft → Rect; hue and LED flash follow; configurator Process B/C/D start values and live long-press stay in sync
- [ ] Fader scrub while a stage button is held cancels mute and process cycle on release (Third-layer edit does not change process)
- [ ] Short press mutes without a click; Shift + long cycles out range (±5 V = 1× flash, 0–10 V = 2× flash)
- [ ] Scene save/load restores amounts, offsets, slews, processes, mutes and Ch0 LFO / input state
- [ ] MIDI CC/NRPN mirrors all four cumulative levels; Ch Map / CC Map of 0 keep base-channel / base-CC-plus-offset behaviour


kosmar:
## Update

Ripppple is now a **signal chain**, not a param-cascade of stage LFOs:

- Ch0: attenuverted CV in, or idle morphing SLFO
- B/C/D: Fold / Soft / Rect on the predecessor (Main=amount, Alt=offset, Third=slew)
- Jacks + MIDI CC = cumulative result
- Gestures (mute, process cycle, range blink, dim pulse, hues) kept

Docs in `docs/apps/46-ripppple/` updated accordingly.

kosmar:
Pushed: 30° process hues (`200→80`) from shared Ch0 start with Manifold.

kosmar:

ArthurGibert:

Ran `ripppple.rs` against the actual submission gate (`pr-scope-check.sh`) — clean, no unsafe code, no panics, no MAX11300/storage bypass, no un-yielding loop — and did a real `cargo build` against current `main` with it dropped in, which passed too. It does have 13 `.unwrap()`/`.expect()` calls without a justification comment, which the gate soft-flags for human judgment rather than blocking (same treatment some existing official apps get) — worth a look before resubmitting, not a blocker.

**About `docs/apps/46-ripppple/manual.{json,md}`**: this one's on us, not you — it was listed as a sanctioned option in `CONTRIBUTING.md`/the scope-check script without anyone actually verifying it was backed by real tooling, and it wasn't: nothing in the configurator reads it (confirmed against #606; its entry doesn't show up on the `/#/manual` page at all). You followed the policy as documented — the policy was wrong. Filed as [#656](https://github.com/ATOVproject/faderpunk/issues/656), and [#657](https://github.com/ATOVproject/faderpunk/pull/657) has already removed it as an option here. So `faderpunk-community-apps` isn't just a better fit organizationally — right now it's the only place this content would actually end up visible anywhere.

kosmar:
Moving this to faderpunk-community-apps as requested: https://github.com/ATOVproject/faderpunk-community-apps/pull/18

Closing here. App ID remapped 46 → 115 (community range).

PR #659, opened 2026-08-23
fix: improve I2C control and panner output

thorinside:
## Category

- [ ] New app
- [ ] App fix
- [x] Firmware core
- [ ] Configurator
- [ ] Protocol (libfp)
- [ ] Docs
- [ ] CI / tooling
- [ ] Other (explain below)

## Type

- [x] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Chore

## Description

Faderpunk currently stores I2C controller changes in a 16-message FIFO. During continuous movement, that queue can fill with positions that are already obsolete. The result can be delayed movement after the fader has stopped, while the final—and most useful—position may be dropped.

This PR replaces that FIFO with one pending value per physical channel. If a channel changes again before its previous value is sent, the newer value replaces it. Different channels remain independent, so continuous controls stay responsive without replaying stale movement.

It also finishes Panner's existing I2C output. Panner previously read a shared value that was initialized to zero and never updated, so it always transmitted zero. It now sends its processed left and right outputs on its two physical channels, including manual pan, autopan, attenuation, mute, and slew.

The documentation now explains the electrical requirements, one-time startup discovery, recognized addresses, and app-specific output behavior. It also documents disting NT compatibility accurately: the NT's follower address is configurable, and setting it to `0x31` lets Faderpunk use the ER-301-compatible `0x11` controller message that the NT supports.

### Scope

The primary category is firmware core. The PR also touches Panner as an existing consumer, `libfp` to keep the coalescing policy host-testable, and the user-facing I2C documentation. Panner continues to access I2C exclusively through the `App<N>` facade; it does not reach into the hardware task directly.

Bus recovery after an electrical fault is not changed here.

## Scope acknowledgement

- [x] I've read [CONTRIBUTING.md](../CONTRIBUTING.md)'s PR Scope Categories table. The related cross-category changes are explained above.

## Verification

- [x] Tested on actual hardware / in the configurator with a live device

Automated checks:

- `cargo fmt --all -- --check`
- `cargo clippy --bin faderpunk --target thumbv8m.main-none-eabihf -- -D warnings`
- `cargo clippy -p libfp -- -D warnings`
- `cargo test --lib -p libfp`
- `pnpm -C configurator lint`
- `pnpm -C configurator build`

Hardware verification with a disting NT configured as an I2C follower at `0x31`:

- Rapid Control movement reaches the latest position without stale catch-up.
- Panner sends both processed outputs.
- Manual pan and autopan modulation update the mapped NT parameters.

### DIFF
--- a/.github/scripts/pr-scope-check.sh
+++ b/.github/scripts/pr-scope-check.sh
@@ -241,7 +241,7 @@ BANNED_PATTERNS=(
   'migrate_fram' 'factory_reset'
   'MAX_CHANNEL' 'MaxCmd' 'MaxSender' 'crate::tasks::max' 'super::tasks::max'
   'APP_MIDI_CHANNEL' 'MIDI_CHANNEL' 'MIDI_USB_PUBSUB' 'MIDI_DIN_PUBSUB' 'crate::tasks::midi'
-  'I2C_LEADER_CHANNEL' 'I2C_FOLLOWER_CHANNEL' 'crate::tasks::i2c'
+  'I2C_LEADER_PUBLISHER' 'I2C_LEADER_CHANNEL' 'I2C_FOLLOWER_CHANNEL' 'crate::tasks::i2c'
   'BUTTON_PRESSED' 'crate::tasks::buttons'
   'CLOCK_PUBSUB' 'TICK_COUNTER' 'CLOCK_IN_CHANNEL' 'TRANSPORT_CMD_CHANNEL' 'SYNC_ENGINE_CHANNEL' 'crate::tasks::clock'
   'GLOBAL_CONFIG_WATCH' 'set_global_config_via_chan' 'crate::tasks::global_config'
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -159,7 +159,9 @@ has unit tests (`cargo test --lib -p libfp`) — run them when you touch it.
 - **Command Channels**: Apps send commands to hardware tasks via async channels:
   - `MAX_CHANNEL` - Control CV jacks (ADC/DAC configuration, set values)
   - `APP_MIDI_CHANNEL` - Send MIDI messages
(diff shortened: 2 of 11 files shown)


ArthurGibert:
Reviewed and tested this on hardware (disting EX, comparable I2C-follower behavior to the NT). Confirmed:
- Rapid, continuous Control fader movement resolves to the final position without visible catch-up (the FIFO-coalescing fix).
- Panner sends both processed L and R outputs over I2C, mapped to two separate controllers on the EX.

Added two small commits on top:
- `fix(i2c): reuse GLOBAL_CHANNELS instead of a duplicate constant` — `PendingFaderUpdates` had introduced its own `I2C_FADER_CHANNELS = 16` next to the existing canonical `GLOBAL_CHANNELS`; folded it into the one constant. Also narrowed the docs' scaling/numbering claim to the ER-301/TXo path, since it doesn't hold for Ansible.
- `docs(i2c): fix stale I2C_LEADER_CHANNEL references` — `AGENTS.md` and the PR-scope CI check's facade-bypass banned-pattern list still named the channel this PR replaces with `I2C_LEADER_PUBLISHER`; updated both (verified against the scope-check's own test suite, all 16 cases still pass).

Also filed two speculative, non-blocking issues for edge cases noticed during review, neither confirmed on hardware and neither should block this:
- #661 — `handle_fader_update` discards `Range`, so bipolar-range outputs may never produce a negative CV value on the ER-301/TXo path.
- #662 — Ansible's `device_port = (chan/4)<<1` can alias two adjacent physical channels (e.g. Panner's L/R) onto the same port for most placements.

No other issues found in the coalescing pipeline itself — traced the publish/signal/drain sequence against `embassy_sync::Signal`'s actual semantics and it holds up under concurrent publishers.

PR #666, opened 2026-08-29
feat: add installable native community apps

thorinside:
## Category

- [ ] New app
- [ ] App fix
- [x] Firmware core
- [ ] Configurator
- [ ] Protocol (libfp)
- [ ] Docs
- [ ] CI / tooling
- [ ] Other (explain below)

This is one firmware feature with necessary protocol, Configurator, tooling, and
documentation pieces. Splitting those pieces would leave each PR unusable on
its own, so **Firmware core** is the primary category.

## Type

- [ ] Fix
- [x] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Chore

### The player problem

Today, trying a community app means building and flashing a different complete
firmware image. That is slow, easy to get wrong, and makes experimentation feel
more permanent than it should.

This PR adds **FPApps**: small native Rust apps that can be installed from the
Configurator while the normal Faderpunk firmware stays in place. The device has
four app slots. Uploading a package fills one slot; removing it frees the slot;
replacing an app can safely stop the running instance, update the slot, and
restore the layout.

There is no separate “FPApps” page. **Installed Apps** lives at the bottom of
the existing **Apps** page. Once installed, an app appears in the normal app
catalogue above it. Its settings and manual also appear in the same places and
with the same formatting as built-in apps.

### What it looks like

An installed community app appears in the normal catalogue:

Sift in the normal Apps catalogue

The four slots stay in a compact section at the bottom of the page:

Installed Apps slots

Manual content is rendered by the existing manual components rather than
being squeezed into the installer table:

Sift manual with normal formatting

### The simple mental model

An `.fpapp` file contains:

1. a short manifest describing the app;
2. its settings, setup notes, and structured manual;
3. native ARM code compiled for one exact Faderpunk firmware ABI.

The firmware exposes a versioned `HostV1` table. The app calls that table for
faders, buttons, LEDs, CV and gate I/O, MIDI, clock, I2C, quantization,
parameters, scenes, persistence, and timers. `fpapp-sdk` provides the friendly
Rust wrapper, so community source keeps using the same `App<N>` programming
model as official apps.

This is native code, not a sandbox or an interpreter. That choice keeps the
wrapper thin and performance predictable, but it also means packages must be
treated like firmware: install only apps whose source and author you trust.

### Compatibility and recovery

Every package carries an exact 32-byte firmware identity and a versioned ABI.
The Configurator and firmware both reject a package built for a different
firmware. That strict rule is intentional: guessing that two native memory
layouts are “close enough” would turn a clear compatibility error into a crash.

Each of the four flash slots is 128 KiB, with up to 124 KiB available for the
package. An app receives a fixed 8 KiB runtime instance. The firmware validates
the container, checksum, metadata, native entry points, ABI identity, and
runtime size before publishing it to the catalogue.

The slots do not use A/B copies. Installation is staged inside the selected
slot, and an interrupted upload leaves that slot empty. The existing firmware
continues to boot, and the player can simply upload the package again.

### Safe lifecycle behavior

The first hardware prototype exposed an important detail: removing bytes from
flash is not the same as stopping a running app. The final lifecycle now works
in this order:

1. evict the app from the active layout;
2. ask its task to stop;
3. wait for an explicit native completion acknowledgement;
4. drain host work and reset its LEDs and jacks;
5. invalidate cached flash descriptors;
6. erase or replace the slot;
7. restore the layout when appropriate.

That ordering is used by Replace, Remove, and Clear All Apps. It prevents stale
code pointers, two copies of an app running at once, and LEDs remaining lit
after an app is gone. Built-in factory apps keep their existing lifecycle; the
new completion handshake is used only where native flash replacement requires
it.

### Configurator behavior

- Upload and replacement progress update in place without repeatedly resizing
  or repainting the page.
- Remove and replace use in-app confirmation instead of browser dialogs.
- An active installed app is stopped automatically when it must be replaced or
  removed; the player no longer has to edit the layout first.
- Device requests are serialized so polling cannot consume another request's
  reply.
- Old firmware replies produce a clear compatibility message instead of
  crashing the Apps page.
- Untrusted package metadata is validated before it can become settings UI.
- Help text says “installed app” or “community app” where the shorter internal
  term “FPApp” would make the interface harder to understand.

### Documentation and author workflow

`docs/fpapp.md` is the normative, versioned format and ABI guide. It explains
the file layout byte by byte, the host service table, task and memory limits,
install protocol, lifecycle, security model, compatibility rules, authoring
workflow, and the evidence used to validate this implementation. The RFCs are
kept as design history; the main document describes what the code actually
does.

The `fpapp` command can build, inspect, and verify packages. It also adapts the
existing community-app source shape, so authors do not maintain a second copy
of their app just to make a package.

## Related pull requests and merge order

1. [#665 — recognize FPApp tooling in scope checks](https://github.com/ATOVproject/faderpunk/pull/665) is the small review-gate prerequisite.
2. This PR adds the firmware, ABI, package builder, Configurator, and documentation.
3. [faderpunk-community-apps#31 — build community apps as FPApps](https://github.com/ATOVproject/faderpunk-community-apps/pull/31) adds the repository-level build command.
4. [faderpunk-community-apps#30 — add Sift threshold sequencer](https://github.com/ATOVproject/faderpunk-community-apps/pull/30) submits Sift in the community repository's exact three-file format.

Sift's community CI build depends on this PR because the dedicated Sift icon
and native runtime do not exist on official `main` yet.

## Scope acknowledgement

- [x] I've read [CONTRIBUTING.md](../CONTRIBUTING.md)'s PR Scope Categories table. The shared pieces are all required parts of one end-to-end firmware feature, and the CI classifier correction is isolated in #665.

### Final head: `60523faa043c02d388afe991d251442164f0f71a`

- [x] `cargo +1.88.0 fmt --all -- --check`
- [x] `cargo +1.88.0 clippy --bin faderpunk --target thumbv8m.main-none-eabihf -- -D warnings`
- [x] `cargo +1.88.0 clippy -p libfp -- -D warnings`
- [x] `cargo +1.88.0 test -p libfp -p fpapp -p fpapp-sdk` — **141 tests passed**
- [x] `pnpm -C configurator lint`
- [x] `pnpm -C configurator build`
- [x] Optimized RP2350 firmware build completed.
- [x] Release link measured at **1,014,388 bytes flash** and **419,480 bytes static RAM**, leaving 558,476 bytes of the firmware partition and 104,808 bytes of SRAM free.
- [x] Rebuilt Heat Pump, Grooves, and Sift against this exact head and verified every package.

Final package SHA-256 values:

| Package | Size | SHA-256 |
|---|---:|---|
| Grooves | 36,401 bytes | `1dc8af7b173fbfd612fa4b2d0fe4da4123a4b017cbd10f6d333bbfc4ed593dcf` |
| Heat Pump | 39,395 bytes | `dc28066cc55d7dcc3d2430852e9705d2bbffe3bdd34cbe667dd6c28e2aaa6ec1` |
| Sift | 26,323 bytes | `65c8a7cc38dd572bc33c73c5cdc6d73d1e4a7670b1fb0ca7bf9bd47164ca95e9` |

### Live hardware and browser

- [x] Tested on actual hardware / in the Configurator with a live device.

The complete player workflow was tested on commit
`82d03d08e2d88ed7d7334bf558259e709ba04754`: install all community apps,
use all four slots, add installed apps to layouts, open their settings and
formatted manuals, replace and remove active apps, clear every slot, confirm
outputs and LEDs shut down, reconnect, and repeat the crash cases that motivated
the lifecycle work. The user completed the final acceptance pass and left Sift
working in slot 1.

The commits after that hardware pass are review-driven guardrails around the
same path: explicit completion acknowledgements, pre-publication runtime-size
validation, metadata validation, sequence-number wraparound, cache invalidation,
and the factory-app exit-signal race fix. They pass the final automated and
release-build checks above, but the exact final head has not been reflashed.
That distinction is recorded here so the evidence is useful rather than
overstated; a maintainer release-candidate flash should repeat the short
replace/remove/clear smoke test.


ArthurGibert:
Really impressive work! This looks like a great solution to making the user submitted apps available to the normal user. 

I'll have a really good look at this and do a bunch of testing. This is a really big PR so it might take a little while but I'm on it!

thorinside:
Yeah, sorry for the size. I'll be responsive to changes if you need.

Cheers,

Neal

ArthurGibert:
Taking over this PR from here per our earlier discussion.

Pushed so far: Arthur's 6 safety fixes (FRAM bounds, poll-count cap, ABI checks, eviction timeouts) plus today's SDK-completeness work — mirrored `App::global_swing`/`current_tick`/`is_clock_running` (#671) into `fpapp-sdk`'s `App<N>`, added `MidiOutput::send_pitch_bend`, fixed missing deps + the edition-2024 `gen` collision in the generated app crates, and shared their `CARGO_TARGET_DIR`. Community app buildability: 15% → 55% (3/20 → 11/20) on the real catalog.

Next up:
- Watchdog + slot quarantine (the safety net for a hung app) — blocking for merge.
- Two known remaining gaps: no `embassy_time` driver for native FPApps, and a ROPI-incompatible jump table in `harmonica`/`vamp` that `-C jump-tables=no` didn't fix. If you've hit that second one before, would take the input.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

ArthurGibert:
Watchdog + slot quarantine is in, and validated on hardware: a deliberately-wedged app trips the watchdog, the correct slot is quarantined, it survives power cycles, and the configurator shows it as Stopped. Reinstalling or removing clears it.

Two bugs in the existing code turned up while testing on device:

- `run_fpapp` is `pool_size = 16` with an 8 KiB `InstanceStorage` in its future — ~128 KiB of the 256 KiB task arena. Spawning hit `task arena is full`. Now derived from `GLOBAL_CHANNELS`.
- `App::current_tick` in fpapp-sdk read the two 32-bit halves without a retry, so a tick landing between them composes a value that never existed. Worst at the `u64::MAX` "not started" sentinel, which every clock start crosses. Now a seqlock-style re-read.

Also worth flagging: slot erase is now split per 4 KiB sector with a watchdog feed between each. `blocking_erase` runs the whole range in a single `in_ram()` call, which pauses Core 1 *and* holds a critical section on Core 0 — so a full-slot erase was a multi-second window where neither core could feed anything. Verified with a 117 KB package installed while another app was running.

The watchdog only arms when a slot holds an app that can actually run, so units with no community apps behave exactly as before.

ArthurGibert:
Status: everything else is done and verified, but there's a RAM regression that blocks merge.

**The problem.** `run_fpapp` is `pool_size = GLOBAL_CHANNELS`, and each future carries a fixed 8,192-byte `InstanceStorage` plus a 3,384-byte `RuntimeContext` — ~11.6 KiB per instance, so the pool reserves **~181 KiB**. Embassy allocates a task pool wholesale on first spawn and never frees it, so spawning a single FPApp costs the full 181 KiB for the rest of that boot.

That starves the built-in apps. A layout Arthur has used live for months (4x control, 2x lfo, ad, tb3po, 2x automator, genseq — six distinct pools) now panics with `task arena is full` once any FPApp has run. It loads fine straight after a factory reset, i.e. when the FPApp pool hasn't been allocated, which is what pins it on this branch rather than on something pre-existing.

Raising the arena isn't a fix: giving built-ins back the 256 KiB they have on `main` while keeping a 16-deep pool needs ~437 KiB of arena, plus the 131 KiB Core 1 stack and ~28 KiB of statics — about 596 KiB against 512 KiB of RAM. Moving `InstanceStorage` into a static doesn't help either; it relocates the cost rather than removing it.

**Being investigated on a branch before merge**, measuring rather than guessing:
- what real apps actually need — `MAX_INSTANCE_BYTES` is a cap, not a measurement, and every instance pays it in full;
- whether the 131 KiB Core 1 stack is really used, since Embassy tasks are futures;
- what a busy built-in layout genuinely consumes.

Any one of those could remove the need to cap how many community apps run at once.

Also worth flagging separately: both spawn paths in `macros.rs` `.unwrap()` the spawn result, so pool exhaustion panics the device rather than leaving the channel empty. Fixing that first, standalone.

Everything else on the branch is green — CI passes all four checks, and the watchdog/quarantine, ABI-contract and context-aliasing work is verified on hardware.

PR #668, opened 2026-08-30
fix: turing mute ignores gate/MIDI outputs

kosmar:
## Summary
- Mute only wrapped the CV pitch path; with Gate Out enabled, gates and MIDI kept firing while the button LED showed muted.
- Gate Out now respects mute (forces gate low while muted), and muting immediately releases a held gate / note-off.

## Test plan
- [ ] Load Turing, enable Gate Out, short-press mute — gate jack stays low and MIDI notes stop
- [ ] Mute in CV mode — no new pitch/MIDI on clock; unmute resumes
- [ ] Mute while a gate is high — gate drops and note-off fires immediately

### DIFF
--- a/faderpunk/src/apps/clkturing.rs
+++ b/faderpunk/src/apps/clkturing.rs
@@ -243,10 +243,9 @@ pub async fn run(
                     * curve.at(storage.query(|s| s.att_saved)) as u32
                     / 4095) as u16;
 
-                let out = quantizer.get_quantized_note(att_reg).await;
-
                 let muted = glob_muted.get();
                 if !muted {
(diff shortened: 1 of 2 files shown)


ArthurGibert:
Also applied the same mute fix to clkturing.rs (Turing+), which has the same underlying bug: `quantizer.get_quantized_note` was called unconditionally every tick regardless of mute state, and muting didn't immediately kill a sustaining MIDI note.

Turing+ has no gate output (CV/MIDI only), so there's no gate-jack/LED-reset branch to fix here — just:
- moved the quantizer call inside `if !muted`
- send `note_off` immediately when mute is toggled on

Verified on hardware via probe-rs flash.

Issue #684, opened 2026-09-12
FPApp: build apps as ropi-rwpi so host access needs no API divergence


ArthurGibert:
Design issue. Supersedes the approaches sketched in #681 and #678 — both are symptoms of one root constraint.

## The constraint

Apps are built `-C relocation-model=ropi` (read-only position independent) and `fpapp.x` allocates only `.text`. `fpapp pack`'s `verify_sections` then rejects any writable allocated section, so **an installed app has no writable statics at all** — confirmed on a real package: one `.text` section (`AXMS`), no `.data`, no `.bss`.

Consequence: **no plain function can reach the host.** There is nowhere to stash the `HostV1` pointer. The embassy-time driver works around this by recovering the host from the waker, which only works where a waker is in hand. A sync method like `Curve::at(&self, u16)` has neither a host argument nor a waker.

Everything below is the same bug wearing different clothes:

- **#681** — `Waveform::at` / `Curve::at` index `libfp` statics, so every package carries its own copy of seven `[u16; 4096]` tables. That's **~70% of each large package** (super-lfo 57,344 of 81,204 B).
- **#678** — `compat::pitch_as_counts` can't resolve custom V/Oct curves, so the same app source produces different CV as a built-in vs. as an FPApp.
- `compat::get_global_config()` is a stub returning zeros for the same reason.

## Why the obvious fixes are unsatisfying

Moving the affected calls onto `App` methods works (that's the #671 pattern) but **splits the API**: factory apps write `Curve::Exponential.at(x)`, community apps write `app.curve_at(...)`. Community app source is supposed to be written exactly like a factory app, with the FPApp conversion happening only at submission. A permanent idiom split defeats that and invites divergence.

Replacing the computable tables with closed forms (verified exact for saw, saw-inv, square, triangle — 32 KB) reduces the symptom without addressing the cause, and leaves sine and the curves duplicated.

## Proposal: `ropi-rwpi`

`rustc` supports `ropi-rwpi` for this target. It adds a static base register for read-write data, which permits writable statics placed in host-provided RAM. The SDK stashes the host pointer at init; from then on *any* function can reach the host, and the API stops needing to change:

```rust
pub fn at(&self, index: usize) -> u16 {
    // identical signature and call sites in both builds
    #[cfg(feature = "fpapp-host")]      { host_table_lookup(self.kind(), index) }
    #[cfg(not(feature = "fpapp-host"))] { WAVEFORM_SINE[index] }
}
```

[...]

Issue #681, opened 2026-09-12
FPApps: waveform/curve tables are ~70% of every large package


ArthurGibert:
## Problem

`libfp::constants` holds seven `[u16; 4096]` lookup tables (five waveforms, two curves) at **8,192 bytes each**. `libfp` is linked into every generated app crate, so each `.fpapp` carries its own full copy of every table it transitively references — while the firmware already contains identical copies.

Measured against the real catalogue:

| App | Native image | Lookup tables | Share |
|---|---|---|---|
| super-lfo | 81,204 B | 57,344 B (7 tables) | **70.6%** |
| ripppple | 68,416 B | 49,152 B (6) | **71.8%** |
| manifold | 70,778 B | 49,152 B (6) | **69.4%** |
| heat-pump | 33,388 B | 16,384 B (2) | **49.1%** |

No app references the tables directly. They arrive transitively through exactly two methods in `libfp/src/lib.rs`:

- `Curve::at(&self, value: u16) -> u16` — indexes `CURVE_EXP` / `CURVE_LOG`
- `Waveform::at(&self, index: usize) -> u16` — indexes the five `WAVEFORM_*` tables

That narrow surface is what makes this fixable without touching the app-facing API: apps keep calling `.at()`.

Note this is *not* an accessibility problem — `libfp` is already linked into app crates. That's precisely the cause: static linking duplicates the data.

## Why it matters

It's the dominant term in package size, and it currently constrains `SLOT_COUNT`. The FPApp flash region is 512 KiB split into fixed-size slots (`SLOT_SIZE = FPAPP_REGION_SIZE / SLOT_COUNT`), and a package must fit entirely within one slot — space in an empty neighbour is unusable. So the largest single package sets the floor:

| Slots | Payload/slot | Apps that don't fit today |
|---|---|---|
| 4 (current) | 126,976 B | none |
| 5 | 100,761 B | none |
| 6 | 83,285 B | super-lfo, manifold |
| 8 | 61,440 B | super-lfo, manifold, ripppple |

With the tables shared, those three drop to roughly 29–35 KB and **every app fits comfortably at 8 slots**, with room to consider more.

## Proposed fix

Serve the tables from the firmware over the host ABI, so `Curve::at` / `Waveform::at` stop carrying their own data in an app build.

[...]

Issue #678, opened 2026-09-12
FPApps: pitch_as_counts silently ignores custom V/Oct curves


ArthurGibert:
Not a blocker for #677 — that PR's helper is correct for what the ABI currently exposes, and documents the limitation. Filing so the gap is tracked rather than lost.

## Problem

`fpapp-sdk`'s `compat::pitch_as_counts` and the firmware's `app::pitch_as_counts` have the same name and signature, and community app source calls it identically — but they behave differently:

- **Firmware** (`faderpunk/src/app.rs:890`) calls `get_global_config().pitch_as_counts(...)`, which resolves a user's Custom V/Oct curve from the live global config.
- **FPApp SDK** calls `Pitch::as_counts(...)`, the plain variant that assumes the standard curve.

So the *same app source* produces different CV output depending on whether it's compiled as a built-in app or packaged as an FPApp. No error, no warning — just slightly wrong pitch for anyone using a calibrated curve.

**Reachable today**: `dit_dah` exposes `Param::VoltPerOct` (`apps/dit_dah.rs:71`) and calls `pitch_as_counts` at line 711. A user selecting one of the four custom curves for that app gets standard-curve output instead.

## Root cause

`GlobalConfig::custom_voct_curves` isn't exposed through `HostV1`, and the SDK's free function has no host handle to fetch it with, so it can only call the curve-less variant.

## Proposed fix

Follows the pattern #671 already established for `global_swing` / `current_tick` / `is_clock_running` (free-function/global access replaced by an `App<N>` method backed by `read_value`):

1. **New value kind**, e.g. `value_kind::CUSTOM_VOCT_CPO = 16`, using the existing `index` argument (0–3) to select the curve. Firmware dispatch is one arm in `fpapp_runtime_services.rs`, alongside the four kinds that already call `get_global_config()`:
   ```rust
   value_kind::CUSTOM_VOCT_CPO => {
       u32::from(get_global_config().custom_voct_curves[index].counts_per_oct)
   }
   ```
   `CustomVoOctCurve` is a single `u16`, and `counts_per_oct == 0` already means "uncalibrated, fall back to Standard (410)", so it maps to `read_value`'s `u32` with no sentinel ambiguity.

2. **Add `App::pitch_as_counts()` and `App::vpo_counts_per_oct()`** as methods on both the real `App<N>` and the SDK facade. The SDK version reads the four curve values and calls `Pitch::as_counts_with_curves(...)`.

[...]

PR #685, opened 2026-09-13
feat(fpapp): build installed apps as ropi-rwpi so they can have writable statics

ArthurGibert:
Implements the loader half of #684. Unblocks #678 and #681 without the API divergence both would otherwise need.

## Problem

Apps were built `-C relocation-model=ropi` and `fpapp.x` allocated only `.text`, so `verify_sections` rejected any writable section and **an installed app had no writable statics at all**. Nothing could hold the `HostV1` pointer, so no plain function could reach the host — not `Curve::at`, not `pitch_as_counts`, not `get_global_config`. The embassy-time driver only escapes this by recovering the host from a waker, which a sync method never has.

Routing those through `App` methods would work, but splits the API: factory apps writing `Curve::Exponential.at(x)` while community apps write `app.curve_at(...)`. Community source is meant to be written exactly like factory source, with FPApp conversion happening only at submission.

## Approach

`ropi-rwpi` addresses read-write data through a static base register, so the firmware hands each instance RAM and points r9 at it. Apps keep **identical signatures and call sites in both builds** — the divergence lives entirely in packaging.

Only zero-initialised statics are supported. `pack` rejects a non-empty `.data` with an explanation rather than silently dropping it, which would hand an app uninitialised statics it believes are initialised. Every app in the catalogue produces neither section today.

## What changed

- `-C relocation-model=ropi-rwpi`; `fpapp.x` gains `.data`/`.bss` at their own base (address 0 collided with `.text`).
- `verify_relocations` accepts `R_ARM_SBREL32` — resolved at link time, no runtime fixup, position-independent by construction, which is the property that check exists to enforce.
- `verify_sections` allows `.data`/`.bss` specifically; anything else writable still fails.
- `pack` reads `.bss` into a new manifest field; manifest key 16, `FPAPP_ABI_MINOR` 1 → 2 (the additive case the rule covers).
- Runtime: per-instance `RwBlock`, refuses apps over the limit, and four `asm!` trampolines setting the static base on entry.

## Verified rather than assumed

**On hardware**: an app carrying a real writable static — a 4-byte atomic bumped on every pitch write, **impossible to build before** — packages with `rw_bytes = 4`, installs on channel 16 and runs normally alongside two factory LFOs. Channel 16 is also the case that used to trigger the task-arena reboot, so it doubles as a regression check.

**The relocation contract**: the linker bakes offset **0** for the `SBREL32`, i.e. relative to the RW segment start, so the firmware's base pointer is used directly as implemented. Had it baked the absolute base instead, the loader would have been silently wrong.

**Emitted code**: disassembly confirms `mov r9, rN` immediately followed by `blx` at each entry.

**Backward compatible**: apps built before this declare `rw_bytes = 0` and exercise the new path without needing any of the new machinery. Existing installed apps kept working across the change.

## Register contract

r9 is callee-saved in the firmware's own code, so a host callback preserves and restores it and the app's base survives the round trip — that direction needs nothing. Only the inbound direction is handled here. `lateout("r9") _` is what makes it safe: without it an input could be allocated into r9 and destroyed by the `mov` before the call.

`clobber_abi("C")` is deliberately not used — it names D16–D31, which this target lacks, and clobbering reserved registers is UB. The AAPCS call-clobbered set is listed explicitly.

## Not included

The payoff — `libfp` resolving tables and curves through the host under a cargo feature — is deliberately a separate change, so the loader can be proved on its own first.

## Gates

`cargo fmt --all -- --check`, firmware clippy, `libfp` clippy, `fpapp` clippy, `cargo test --lib -p libfp` (133 pass, including the golden package test updated for the new wire format).


ArthurGibert:
## Hardware verification — passed

Built a dedicated probe app rather than relying on judging a normal app's output, since "it didn't crash" isn't proof and the earlier test asked the maintainer to assess behaviour they had no reference for.

The probe (`rw_probe`, 8 bytes of writable statics, packaged with `rw_bytes = 8`) checks two independent things and reports the answer as colour:

1. Writes a magic value into a static once, re-reads it on every pass — catches a wrong base pointer.
2. Increments a counter across separate polls — catches statics that don't persist between entries into app code.

**Result on device: top LED solid green, bottom LED blinking green.** Both checks pass — the magic survived, and the counter is advancing, with the blink *being* the increment.

Also confirmed earlier in the same session:

- A patched Dit Dah carrying a real writable static installed on **channel 16** and ran alongside two factory LFOs. Channel 16 is the case that used to trigger the task-arena reboot, so it doubles as a regression check.
- Existing installed apps, which predate this and declare `rw_bytes = 0`, kept working unchanged across the firmware swap.
- Clean boot log: `boot: watchdog timeout=false scratch0=0x00000000`.

Neither app could have been *built* before this change — a writable static compiles to an absolute address `fpapp pack` rejects.

One note for reviewers: an SWD flash still half-fails and drops the unit to BOOTSEL when the watchdog is armed. That is the pre-existing documented interaction (the probe's erase runs on the target core with nothing feeding the watchdog), reproduced again here and unrelated to this change.

